PR #115, opened 2026-06-06
eval() support + co-locate SharpTS.dll only when needed (#107)

### LINKED ISSUE #107: Test262 #8: eval(...) rejected in property-accessors and elsewhere (~10+ tests, broader impact)
## Motivation

Surfaced during the #69 property-accessors rollout (commit 2f92515). Test262 uses `eval("…")` to verify parser-level whitespace tolerance:

```js
if (eval("Number	.	POSITIVE_INFINITY") !== Number.POSITIVE_INFINITY) {
  throw new Test262Error('...');
}
```

In SharpTS this buckets as `TypeCheckError`, suggesting our type checker either:
1. Doesn't have `eval` typed in lib.d.ts.
2. Special-cases `eval` and rejects it under TS strict.
3. Treats the return as something other than `any`, causing the `!==` comparison to fail type-check.

## Sample failing tests

`test/language/expressions/property-accessors/S11.2.1_A1.1.js` and siblings (~10 tests in property-accessors).

`eval` also lurks in many other Test262 directories we haven't yet rolled out; addressing this now lifts a categorical floor before those rollouts.

## Impact

| Folder | Tests using `eval()` |
|--------|---------------------:|
| `language/expressions/property-accessors` | ~10 |
| `language/expressions/{this, comma, ...}` (not yet rolled out) | unknown |
| `language/statements/{block, with, ...}` (not yet rolled out) | unknown |

The PA hit alone is small; the cross-cutting impact on future rollouts (priority-2/3 folders in #69) is much larger.

## Suggested approach

Two open questions to investigate:
1. **Is `eval` parsed and typed at all?** Spot-check `Parsing/Parser.*.cs` and `TypeSystem/*` — search for `"eval"` references. If `eval` is special-cased to reject, lift the rejection in non-strict-mode contexts at minimum.
2. **Does SharpTS support `eval` at runtime?** True dynamic eval requires re-running the parser/type-checker/interpreter against a string. Compile mode is harder (eval would need JIT). If we don't and won't support runtime `eval`:
   - Type the global `eval` as `(s: string) => any` so calls *type-check* and run as a no-op stub returning `undefined`.
   - That makes Test262 tests fail at runtime with a deterministic outcome (RuntimeError "eval not supported") rather than `TypeCheckError`, surfacing the gap.

Decision: do we support eval at all? If not, the lib.d.ts stub is the right fix.

## Acceptance

- The ~10 property-accessor tests using `eval()` move from `TypeCheckError` to either `Pass` (if eval works) or `RuntimeError: eval not supported` (if we choose the stub route).
- A clear documented stance on `eval()` in CLAUDE.md / STATUS.md so future rollouts don't re-investigate this.

## Related

Part of #69. Pre-condition for cleaner rollouts of `language/expressions/{this, comma, ...}` and `language/statements/*` (priority-2/3 in #69).

nickna:
## Summary

Two related changes:

1. **`eval()` support (closes #107)** — real eval in both execution modes.
2. **Runtime-dependency signal + auto-copy** — co-locate `SharpTS.dll` with compiled output only when a feature actually needs it at runtime.

## 1. eval() (#107)

`eval` is now registered as a global typed `(s: string) => any` (identifier + call paths), removing the categorical `TypeCheckError` floor that blocked any program referencing eval.

- **Interpreted:** direct eval — `Interpreter.Eval()` lexes/parses/interprets the source against the caller's scope chain and returns the completion value. Not type-checked (matches tsc); non-string args returned unchanged (ECMA-262 §19.2.1); parse failures throw `SyntaxError`.
- **Compiled:** indirect eval via `EvalBridge` in the SharpTS runtime, invoked through the `Type.GetType("…,SharpTS")` reflection pattern so the output DLL keeps **no hard SharpTS reference**. Global builtins resolve; compiled locals are not visible (no live interpreter behind native IL). Degrades to a deterministic "eval not supported" throw when SharpTS.dll is absent.

### Acceptance verified
The issue's criterion — the property-accessor eval tests move off `TypeCheckError` — is confirmed: `test/language/expressions/property-accessors/S11.2.1_A1.1.js` now **passes all 9 Unicode-whitespace checks** (TAB, VT, FF, SP, NBSP, LF, CR, LS, PS) in **both** interpreted and compiled modes (`TypeCheckError → Pass`).

## 2. Co-locate SharpTS.dll only when needed

Some features (eval, Proxy, Intl, vm, dns, `@DotNetType` dynamic events) emit a `Type.GetType("…,SharpTS")` path whose *normal* execution needs SharpTS.dll present. A byte-scan for `,SharpTS` is too coarse — the unconditional `process` helper bakes that string into every DLL, yet a trivial program runs fine standalone. So this adds a **semantic** signal:

- `EmittedRuntime.RequiredSharpTSRuntimeReasons` + `RequireSharpTSRuntime(reason)`, surfaced via `ILCompiler.RequiredSharpTSRuntimeReasons` (one set per compilation).
- Recorded at the gates/call sites that emit a *reachable* late-binding path; **not** recorded for pure-BCL features (zlib, child_process, JSON) or unconditional graceful-fallback plumbing (`process`).
- `Program.cs` copies `SharpTS.dll` next to the output after `Save` **only when** the reason set is non-empty — programs using none of these stay fully standalone.
- New `--compile … --standalone` flag suppresses the copy (those features then throw their clear "not supported" error at runtime).

### Verified end-to-end
- Trivial program → **no copy**.
- `eval` program → `Copied SharpTS.dll … required at runtime by: eval()`; co-located DLL runs.
- `Proxy` program → copy with reason `Proxy`; runs.
- `--standalone` → note printed, **no copy**.

## Tests
- `EvalTests` — both modes (whitespace tolerance, completion values, non-string passthrough, var decls, builtin calls) + an interpreter-only direct-eval local-capture case.
- `RuntimeDependencySignalTests` — trivial = empty reasons; eval/Proxy/Intl flagged; eval does not false-positive other features.
- Full suite: **10335 passed**; only the 2 pre-existing `PackagingTests.Pack_WithoutPackageJson_*` failures remain (confirmed failing at baseline `main`, unrelated). Standalone-DLL guard green.

## Known limitations / follow-ups (not in scope here)
- Compiled eval is indirect-only (can't see compiled locals) — by design.
- `AbortSignal.any` late-binds but is gated only at the coarse `UsesAbortController` level, so it's left unflagged to avoid penalizing common AbortController + fetch use.
- The `-t exe` and NuGet packaging paths don't yet consume the runtime-dependency signal.

Closes #107.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -136,6 +136,8 @@ The same applies to `PropertyDescriptorStore`, `ObjectBuiltIns`, and any other S
 
 **Why:** When emitting IL with `typeof(X).GetMethod(...)`, the method token references the SharpTS assembly directly. This creates a hard dependency. The reflection pattern emits IL that does `Type.GetType("..., SharpTS")` at runtime, allowing graceful degradation if SharpTS isn't present.
 
+**Soft-dependency signal + auto-copy:** Some features (eval, Proxy, Intl, vm, dns, `@DotNetType` dynamic events) emit a `Type.GetType("…, SharpTS")` path whose *normal* execution needs SharpTS.dll present. When such a path is emitted, the emit site records a reason via `EmittedRuntime.RequireSharpTSRuntime(reason)`, surfaced through `ILCompiler.RequiredSharpTSRuntimeReasons`. After `Save`, the CLI (`Program.cs` `CopySharpTSRuntimeIfNeeded`) co-locates SharpTS.dll with the output **only when** that set is non-empty — programs using none of these stay fully standalone. `--compile … --standalone` suppresses the copy (those features then throw a clear "not supported" error at runtime). Do NOT record reasons for pure-BCL features (zlib, child_process, JSON) or unconditional graceful-fallback plumbing (`process`) — only paths a normal program actually reaches.
+
 ### Error Handling Conventions
 
 - Type errors: "Type Error:" prefix
--- a/Cli/CommandLineParser.cs
+++ b/Cli/CommandLineParser.cs
@@ -52,7 +52,8 @@ public record CompileOptions(
     bool MsBuildErrors = false,
     bool QuietMode = false,
     IReadOnlyList<string>? References = null,
-    BundlerMode Bundler = BundlerMode.Auto
+    BundlerMode Bundler = BundlerMode.Auto,
+    bool Standalone = false
 )
 {
     public IReadOnlyList<string> References { get; init; } = References ?? [];
@@ -277,6 +278,7 @@ public class CommandLineParser
         bool verifyIL = false;
         bool msbuildErrors = false;
         bool quietMode = false;
+        bool standalone = false;
         string? sdkPath = null;
         BundlerMode bundlerMode = BundlerMode.Auto;
 
@@ -376,6 +378,10 @@ public class CommandLineParser
             {
                 quietMode = true;
             }
+            else if (args[i] == "--standalone")
+            {
+                standalone = true;
+            }
             else if (args[i] == "--pack")
             {
                 pack = true;
@@ -415,7 +421,8 @@ public class CommandLineParser
             MsBuildErrors: msbuildErrors,
             QuietMode: quietMode,
             References: references,
-            Bundler: bundlerMode
+            Bundler: bundlerMode,
+            Standalone: standalone
         );
 
         var packOptions = new PackOptions(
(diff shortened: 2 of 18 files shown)


nickna:
Follow-ups for the runtime-copy mechanism (out of scope for this PR), now tracked:
- #116 — flag `AbortSignal.any` precisely (not all of `UsesAbortController`)
- #117 — consume `RequiredSharpTSRuntimeReasons` in the `-t exe` and NuGet `--pack` paths

Issue #188, opened 2026-06-10
TS Conformance: contextual signature instantiation for generic signatures (~20-test Fail cluster)


nickna:
Part of #80. Continuation of the signature-assignability track (#169 was Step 1).

## Problem
The largest remaining Fail cluster in the conformance baseline (~20 of 42 Fails) is function/construct-signature assignability involving **generic signatures**:
- `assignmentCompatWithCallSignatures3–6` + `...WithOptionalParameters`
- `assignmentCompatWithConstructSignatures3–6` + `...WithOptionalParameters`
- `assignmentCompatWithGenericCallSignatures{2,4}` + `...WithOptionalParameters`
- `callSignatureAssignabilityInInheritance{,3,6}`, `constructSignatureAssignabilityInInheritance{,3,6}`
- `covariantCallbacks`

Root cause (investigated 2026-06-08): generic signatures parse to `TypeInfo.GenericFunction`, which historically had no case in `IsCompatibleCore`. #169 (Step 1) unified Function/GenericFunction behind `SignatureRelatedTo` and handles generic-source → concrete-target via naive inference. What's missing is faithful **contextual signature instantiation** (tsc's `signatureRelatedTo`).

## Plan
**Step 2 — faithful contextual instantiation** (branch `feature/generic-signature-instantiation` has groundwork: gated GenericFunction parsing of generic function-type annotations + variance-aware candidate inference, unit suite green, 0 conformance drift):
- [ ] bidirectional **both-generic** relating (currently gated to lenient — comparing distinct opaque type params falsely rejects; blanket leniency breaks `callSignatures5`)
- [ ] overloaded-target handling
- [ ] nested-function-type handling
- [ ] un-gate GenericFunction annotation parsing once both-generic relating is real

**Step 3 — re-land TS2430** (interface-extends member compatibility) calling `SignatureRelatedTo` → the six `*AssignabilityInInheritance*` tests. Previous attempt reverted because raw `IsCompatible` on generic members gave false positives (regressed `constructSignatureAssignabilityInInheritance{2,5}`). Must report-and-continue (not throw) so module-nested sibling interfaces still get checked; line attribution via `Stmt.Interface` → `Name.Line`.

**Step 4 — parameter bivariance fix** (`RelateFunctionShapes` is one-directional) → `covariantCallbacks`. Highest blast radius; do last, measured separately.

## Known inference subtleties (from the reverted first attempt)
[...]

Issue #80, opened 2026-04-24
Epic: Add TypeScript conformance test runner (microsoft/TypeScript corpus)


nickna:
## Context

We already have a Test262 runner (`SharpTS.Test262/`) that pins us against the canonical ECMA-262 / JavaScript spec. There is no TC39-equivalent for TypeScript — TypeScript's written spec was deprecated in 2014, and `tsc` is the de-facto reference. The closest thing to a canonical TS conformance corpus is `microsoft/TypeScript/tests/cases/conformance/` (~10–15k `.ts` files with `*.errors.txt` / `*.js` / `*.types` / `*.symbols` baseline sidecars). Microsoft's own native rewrite (`microsoft/typescript-go`) is measured against this exact corpus.

This epic adds a parallel runner project, `SharpTS.TypeScriptConformance/`, that mirrors the Test262 setup so prior knowledge transfers directly.

## Outcome

- Vendored `microsoft/TypeScript` checkout under `external/typescript/`
- New `SharpTS.TypeScriptConformance/` project (xUnit, not in `SharpTS.sln`)
- Phase-1 errors-baseline runner: diff our type checker's diagnostics against TS's `*.errors.txt`, comparing on diagnostic *code* + approximate location (not exact wording)
- Committed initial baseline on a small subset; bucket-diff harness identical to Test262
- CI job that runs on PRs and gates on regressions

## Why phased

1. **Errors-only** is the highest signal-to-effort ratio — it directly exercises the type system, the part that distinguishes TS from JS.
2. **Emit conformance** (compare against `*.js`) is lower priority for SharpTS since we compile to IL, not JS — useful only for transpile paths we keep.
3. **Types/symbols** introspection requires exposing per-node type/symbol data from the checker; high effort, high value if/when we want LSP/hover parity.

## Caveats

- `microsoft/TypeScript` is Apache-2.0 — submodule + LICENSE preservation is fine.
- TS diagnostic *wording* changes between versions; match on codes (`TS2322`) and approximate location, not full text. Pin to a specific TS tag.
- Some tests need `lib.es*.d.ts` from the same TS checkout — must resolve `lib` against the vendored copy.
- Most conformance tests are *negative* — runner compares error sets, not exit codes.

## Reference

- `SharpTS.Test262/Test262Runner.cs` — bucket model + thread/timeout/memory watchdog
- `SharpTS.Test262/Test262BaselineDiffer.cs` — committed-baseline diff with hard-fail on regressions/new-passes
[...]

PR #201, opened 2026-06-10
feat: contextual signature instantiation + TS2430 (#188 Steps 2-4)

nickna:
Part of #80; implements Step 2 of #188. Builds on #169's `SignatureRelatedTo` unification.

## What
Replaces the both-generic leniency gate with faithful relating, mirroring tsc's `signatureRelatedTo`/`signaturesRelatedTo`:

- **Both-generic relating**: the source signature is instantiated against the target with the target's type parameters rigid. Constraint-violating inferences are **clamped** to the constraint (tsc rule); conflicting candidates (no common subtype, e.g. `<T>(x:{a:T;b:T})` against `(x:{a:U;b:V})`) leave the parameter **rigid** so the relation fails.
- **Inference priority**: parameter-position candidates beat return-position ones (relating `<T>(x:T)=>T` to `(x:U)=>void` infers `T := U`, not `U | void`).
- **Alpha-renaming**: bound signature type params get fresh side-distinct names before relating, so `<T>() => T` can't capture a free outer `T` (type-param identity is name-based in this checker). Constraint chains are re-resolved per pass.
- **Erased relating for overloads**: when either side has multiple signatures, pairings relate with type params erased to `any`; only single-vs-single gets full instantiation — exactly tsc's `signaturesRelatedTo` split.
- **Unified call/construct signature matching**: the `IsGeneric → skip` gates are gone; interface and object-literal signature lists route through `SignatureRelatedTo`. This fixes the **inverted construct-signature direction** (generic targets were vacuously satisfied while generic sources were excluded).
- **Bivariant parameter relation** (tsc default; `strictFunctionTypes` is a later opt-in tightening) — Step 4 of the #188 plan, pulled in because the cluster baselines predate strictFunctionTypes.

## Latent bugs surfaced and fixed
- `Function`/`GenericFunction`/`CallSignature`/`ConstructorSignature` `ToString` omitted optionality/rest markers → ToString-keyed compat-cache collisions (same class as cb672dee). `(x?: number) => 1` and `(x: number) => 1` shared a verdict.
- `SplitObjectMembers` counted the `>` of `=>` as a closing bracket — two call signatures in an object type fused into one curried mess.
- `Array<T>`/`ReadonlyArray<T>` annotations resolved to **any** (no built-in generic case).
- `(y: Array<Base>) => void` mis-routed to `ParseGenericTypeReference` → any.
- Interface call/construct signature type params were never in scope while member types resolved → `<T>(x: T): T` collapsed to any and related vacuously.

## Result
- Conformance subset: **12 Fail → Pass**, 0 regressions (Pass 33 → 45, Fail 41 → 29). The entire `assignmentCompatWith{Call,Construct}Signatures{3,4,5,6,WithOptionalParameters}` + `GenericCallSignatures{2,WithOptionalParameters}` cluster now passes.
- Still Fail (expected): `genericCallSignatures4`, `covariantCallbacks` — both need structural variance for generic interface instantiations; `*AssignabilityInInheritance*` await TS2430 (Step 3).
- Full unit suite: 10493 passed, 2 known pre-existing packaging failures. 16 new unit tests.


nickna:
Pushed Step 3 on top: **TS2430 interface-extends member compatibility** (`a395d776`).

Every member a derived interface redeclares must be assignable to the corresponding member of each extended interface. Key design points:
- Validation runs **after** the interface is defined, so an incorrect extension still leaves the type resolvable (no cascading unknown-type errors).
- Interfaces inside a namespace are checked in the namespace's non-recovering collection pass, so in recovery mode the diagnostic is **recorded directly** (report-and-continue) — sibling interfaces in the same module still get checked, which the conformance tests require (two erroring interfaces per module).
- The earlier attempt at this check was reverted for generic-member false positives; with contextual signature instantiation relating generic members faithfully, it lands clean — **zero new failures across the 10.5k-test unit suite**.

Cumulative for this PR: **18 Fail → Pass** on the conformance subset (Pass 33 → 51, Fail 41 → 23), 0 regressions. The PR now covers #188 Steps 2, 3, and 4.

PR #204, opened 2026-06-10
Bugfix batch: #190-#197, #199

### LINKED ISSUE #190: Arithmetic on undefined crashes (InvalidCastException) instead of producing NaN — both compiled and interpreter
## SymptomPer ECMA-262, `ToNumber(undefined)` is `NaN`, so `undefined * 2` evaluates to `NaN`. SharpTS instead throws a host-level cast error in **both** execution modes:```tsconst u: any = undefined;console.log("direct=" + (u * 2));   // expected: direct=NaN```**Compiled** (`--compile`, then `dotnet out.dll`) — unhandled process crash:```Unhandled exception. System.InvalidCastException: Unable to cast object of type ''$Undefined'' to type ''System.IConvertible''.   at System.Convert.ToDouble(Object value)   at $Program.Main()```**Interpreter**:```Runtime Error: Unable to cast object of type ''SharpTS.Runtime.Types.SharpTSUndefined'' to type ''System.Double''.```## Notes- Only reachable through `any`-typed values (the type checker rejects the static case), so this is the dynamic/untyped path — likely overlaps with the known Test262 dynamic-path gaps.- Both modes are wrong relative to the spec, but the compiled mode is the worse failure: an uncatchable-looking `InvalidCastException` process crash rather than a guest-level error. Same crash signature also surfaces when an `any`-typed callback does arithmetic on `undefined` array elements (see #191 for how those elements arise).- If strict-over-spec behavior here is intentional, the compiled path should at least fail as a proper guest `TypeError` rather than `System.Convert.ToDouble` throwing raw.Found during the C#-merge feasibility spike while building a semantics-parity corpus for `Array.prototype.map`.

### LINKED ISSUE #191: Array literal elision creates undefined elements instead of holes ([1, , 3])
## Symptom

Elided positions in array literals are materialized as `undefined` elements rather than holes, in **both** compiled and interpreter modes (the two modes agree with each other, so this is an engine-wide spec deviation, not a mode divergence):

```ts
const sp = [1, , 3] as any[];
console.log(1 in sp);                       // actual: true     expected: false
let calls = 0;
sp.map((x, i) => { calls++; return 0; });
console.log(calls);                          // actual: 3        expected: 2
let fe = 0;
sp.forEach((x, i) => { fe++; });
console.log(fe);                             // actual: 3        expected: 2
console.log(sp.join(","));                  // "1,,3" (correct, coincidentally — join renders undefined as empty)
```

Per ECMA-262 (ArrayLiteral / Elision), `[1, , 3]` has `length 3` but **no property at index 1**: `1 in sp` is `false`, and `map`/`forEach` skip the callback for that index (map preserves the hole in its output).

## Notes

- The hole machinery itself exists and works: `$ArrayHole` is emitted, and e.g. `map` correctly preserves holes that arise when a callback truncates the array mid-iteration (verified). The gap is only that **array-literal elision never produces `$ArrayHole`** — the literal lowering stores `undefined` instead.
- Downstream effect: an `any`-typed callback doing arithmetic on the elided element hits the `undefined`-arithmetic crash in #190 instead of being skipped per spec.
- Likely accounts for a slice of the known Test262 dynamic-path failures around sparse arrays.

Found during the C#-merge feasibility spike while building a semantics-parity corpus for `Array.prototype.map`.

### LINKED ISSUE #192: net: connecting to a missing named pipe burns the full 5s timeout before ENOENT (both modes)
## Symptom

`net.createConnection(pipePath)` to a **nonexistent** Windows named pipe takes a flat ~5 seconds before the `error` event fires. Node.js raises `ENOENT` immediately.

Found via test-suite profiling: `IpcSocket_ConnectErrorCode` costs 5.0-6.3s in *both* modes (one of the slowest tests in the suite), and the cost is intrinsic, not contention.

## Root cause

Both runtimes connect with a fixed 5s timeout:

- Interpreted: `Runtime/Types/SharpTSSocket.cs` (~line 216): `await pipeClient.ConnectAsync(5000)`
- Compiled: `Compilation/RuntimeEmitter.TSNetSocket.cs` (~line 638): `pipeClient.Connect(5000)`

`NamedPipeClientStream.Connect(timeout)` does not distinguish "pipe does not exist" from "pipe busy": when the pipe is missing it **retries CreateFile in a loop until the timeout expires**, then throws `TimeoutException`. So the ENOENT case - which Node reports instantly - always costs the full budget.

## Suggested fix

Pre-check pipe existence (e.g. `Directory.EnumerateFiles(@"\\.\pipe\")` contains the target name) and surface `ENOENT` immediately when missing; keep the 5s wait only for the exists-but-busy case (which is what the timeout is actually for). Apply to both the interpreted socket and the compiled emitter path.

Note: avoid `File.Exists(@"\\.\pipe\name")` - probing a pipe path with CreateFile-based APIs can consume a pipe instance / interfere with WaitNamedPipe semantics; enumeration is the safe probe.

### LINKED ISSUE #193: fs.watchFile (compiled): options object is silently ignored - interval always falls back to 5007ms
## Symptom

In compiled mode, `fs.watchFile(path, { interval: 100 }, cb)` polls at the default **5007ms**, ignoring the requested interval. Interpreted mode honors it.

Found via test-suite profiling: `WatchFile_DetectsStatChange(mode: Compiled)` takes 6.7s vs **0.42s** interpreted - the compiled watcher simply does not see the `interval` option.

## Root cause

`Compilation/RuntimeEmitter.FsWatcher.cs` (~lines 383-411, `FsWatchFile`): the options argument is only inspected when it implements `$IHasFields`:

```
il.Emit(OpCodes.Ldarg_1);
il.Emit(OpCodes.Isinst, runtime.IHasFieldsInterface);
il.Emit(OpCodes.Brfalse, afterOptionsLabel);   // <- falls through to 5007 default
```

But compiled **object literals are `Dictionary<string, object?>`**, which does not implement `$IHasFields` - so any options bag passed as a literal (the normal calling convention) is skipped entirely. This affects every compiled program using `fs.watchFile` options, not just tests; `persistent` handling has the same gap.

## Suggested fix

Extend the option-extraction IL to also handle `IDictionary<string, object?>` (TryGetValue on "interval"/"persistent"), mirroring how other emitters treat literal objects. Audit sibling emitters for the same `$IHasFields`-only assumption.

### LINKED ISSUE #194: Interpreter.ScheduleTimer: cross-thread timers with delay > 10ms never wake a blocked event loop (worst-case latency up to 60s)
## Symptom

A timer scheduled from a non-event-loop thread with `delayMs > 10` does not wake an event loop that is already blocked in its wait. The loop only notices the timer when something *else* wakes it, or when its current `TryTake` timeout expires - which is computed as `GetNextTimerTimeout()` **before** the new timer existed, i.e. up to **60 seconds** when no other timers were pending.

## Root cause

`Execution/Interpreter.cs` (~lines 420-437):

```csharp
internal VirtualTimer ScheduleTimer(int delayMs, int intervalMs, Action callback, bool isInterval)
{
    ...enqueue into _virtualTimerQueue...
    // Wake the event loop if the timer fires soon (within 10ms)
    if (delayMs <= 10)
    {
        WakeEventLoop();
    }
    return timer;
}
```

The `<= 10` guard means a 100ms timer scheduled by a thread-pool completion (the pattern used by `SharpTSTlsSocket`, DNS, crypto callbacks, etc. to hop work back onto the loop) can sleep for the remainder of the loop''s current wait window. `RunEventLoop` blocks in `_callbackQueue.TryTake(out action, (int)GetNextTimerTimeout().TotalMilliseconds, ...)` and `GetNextTimerTimeout()` returns 60s when the timer queue was empty at computation time.

## Why it has gone unnoticed

Most runtime code uses `ScheduleTimer(0, ...)` (which does wake) or schedules from the loop thread itself (where the next iteration recomputes the timeout). The bug only bites cross-thread scheduling with delays in the 11ms-60s range.

## Suggested fix

Wake unconditionally, or wake when the new timer''s due time is earlier than the loop''s current wait deadline. The no-op-callback wake (`WakeEventLoop`) is cheap; the 10ms heuristic saves little and costs correctness.

Found while investigating interp event-loop latency for #194-adjacent work (test-suite profiling).

### LINKED ISSUE #195: Promise.prototype.then conformance: onFulfilled throw invokes same-then onRejected (compiled); SpeciesConstructor lookup not implemented
Two related spec gaps in `Promise.prototype.then`, found during Test262 regen work (PR #187).

## 1. Compiled: a throwing onFulfilled invokes the SAME then''s onRejected

```ts
Promise.resolve(1).then(function() { throw new Error("boom"); },
                        function(e) { console.log("onRejected called with: " + e.message); });
```

- Node: onRejected is **not** called (it only handles rejection of the *input* promise); the then-result promise rejects with "boom".
- SharpTS compiled: prints `onRejected called with: boom`.

Cause: in `Compilation/RuntimeEmitter.Promises.Then.cs`, `$PromiseThen_SM.MoveNext` wraps both the input-promise `GetResult()` **and** the `InvokeCallback(onFulfilled, ...)` in one try; the catch dispatches to onRejected. So "input promise rejected" and "onFulfilled threw" are conflated. Per ECMA-262 §27.2.5.4 they must be distinct: a throwing handler rejects the output promise, full stop. (PR #187 hardened this catch path against process-killing escapes - 6650ffca - but the dispatch conflation remains.)

Interpreted mode does not invoke onRejected (it surfaces the throw as a top-level runtime error - also not spec-shaped, but a separate, milder divergence).

## 2. SpeciesConstructor lookup not implemented

`then` must read `promise.constructor` / `Symbol.species` to build the result capability (NewPromiseCapability). SharpTS skips this entirely, so e.g. `test/built-ins/Promise/prototype/then/ctor-poisoned.js` (a poisoned `constructor` getter must make `then` throw synchronously) fails. This test is baselined Fail as of PR #187 - its previous Pass was a false positive that depended on the crash bug fixed there (the escaping handler throw happened to satisfy `assert.throws`).

## Suggested approach

Restructure the then state machine: try around `GetResult` only -> branch on fulfilled/rejected -> invoke the selected handler under its own guard (the nested-guard plumbing from 6650ffca is reusable). SpeciesConstructor support is a larger piece and may be worth splitting out if `then` is restructured first.

### LINKED ISSUE #196: Promise.race([]) settles immediately - spec requires it to remain forever pending
## Symptom

`Promise.race([])` (empty iterable) settles immediately in SharpTS. Per spec it must return a promise that **never settles** - there are no competitors to race.

Observable via Test262 `test/built-ins/Promise/race/S25.4.4.3_A5.1_T1.js`:

```js
var p = Promise.race([]);
p.then(function() { throw new Test262Error("Never settles."); },
       function() { throw new Test262Error("Never settles."); }).then($DONE, $DONE);
```

Both handlers must never run; SharpTS settles `p`, so one runs and throws.

## Status

Pre-existing bug (test has never passed). As of PR #187 it is baselined interpreted=`RuntimeError`, compiled=`Fail` - the bucket shifted from Fail to RuntimeError in interp because the never-settling-promise hang fixes changed *where* the bogus settlement is observed (it now lands on a thread-pool continuation after the top-level wait escapes). The root cause is unchanged: race over an empty iterable resolves/rejects when it should leave the capability untouched.

Likely the same family as Promise.all/allSettled/any empty-iterable edge handling - worth checking those while in there (`all([])` must resolve with `[]`, `any([])` must reject with AggregateError - only `race([])` stays pending).

### LINKED ISSUE #197: Test262 harness: SHARPTS_TEST262_TRACE=1 + worker recycle = sharing violation that fails the run
## Symptom

With `SHARPTS_TEST262_TRACE=1`, any regen where a worker slot recycles (memory ceiling, crash, watchdog) fails the compiled fact with:

```
System.AggregateException : ... The process cannot access the file ''...\sharpts_trace_...\worker_00.log'' because it is being used by another process. (xN)
```

The baseline diff never runs, so traced regens cannot complete on any run long enough to recycle a worker. Observed on the 2026-06-10 traced regen (all 6 slots affected).

## Root cause

`SharpTS.Test262/BatchedSubprocessRunner.cs`: each slot''s respawn loop constructs a new `PersistentWorker` with the **same trace path** (`worker_{id:D2}.log`, append mode), but the previous worker''s parent-side `WorkerTrace` `StreamWriter` is only disposed at the end of `RunAll` (workers are collected into `slotDisposables` and disposed in the `finally`). Second open of the same file while the first writer is live -> `IOException` -> slot task faults -> `Task.WaitAll` throws.

## Suggested fix

Any of: (a) unique file per spawn (`worker_{id:D2}_{spawnIndex}.log`), (b) open with `FileShare.ReadWrite`, or (c) dispose the prior worker before constructing its replacement in the slot loop. (a) is simplest and keeps per-process logs separable.

Trace mode only - normal regens are unaffected (`WorkerTrace` writer is null without the env var).

### LINKED ISSUE #199: PackagingTests: Pack_WithoutPackageJson_RequiresOverrides / _SucceedsWithOverrides failing on main
Both tests fail on current main (verified on ae02c291 via stash-and-run during unrelated work, and consistently since):

```
SharpTS.Tests.PackagingTests.IntegrationTests.Pack_WithoutPackageJson_RequiresOverrides
  Assert.Contains() Failure: Sub-string not found
  String:    "Compiled to C:\\Users\\...\\AppD..."
  Not found: "package.json"

SharpTS.Tests.PackagingTests.IntegrationTests.Pack_WithoutPackageJson_SucceedsWithOverrides
  (fails at IntegrationTests.cs:524)
```

`Pack_WithoutPackageJson_RequiresOverrides` expects the pack flow without a `package.json` to error mentioning the missing manifest/required overrides; instead the CLI output begins with "Compiled to ..." - i.e. the validation either no longer fires or fires after/without the expected message. The sibling override-success test fails alongside it, suggesting one regression in the pack argument/manifest validation path rather than two independent breaks.

These are the only 2 failures in an otherwise green 10,484-test suite, so they also add permanent noise to every full-suite run (exit code 1 masks new failures).

Not yet bisected - first step is `git bisect` against the pack CLI path or a look at recent `Packaging/` / CLI argument changes.

nickna:
Fixes #190, fixes #191, fixes #192, fixes #193, fixes #194, fixes #195, fixes #196, fixes #197, fixes #199.

One commit per bug, in dependency-friendly order:

## Semantics (both modes)

- **#190 — arithmetic on undefined → NaN.** Interpreter binary slow paths now use ECMA-262 ToNumber coercion instead of raw `(double)` casts; compiled `EnsureDouble`/unary/increment emitters call `$Runtime.ConvertToNumber` instead of `Convert.ToDouble` (which threw on `$Undefined` and crashed the process). `ConvertToNumber` gained an `IConvertible` fallback so boxed non-double numerics from `@DotNetType` interop keep converting. Bonus: the type checker typed `-x`/`~x` on `any` as bigint (`IsBigInt(any)` is true via `IsTypeOfKind`), which made any enclosing binary fail at compile time — now typed `number` per tsc.
- **#191 — `[1, , 3]` creates true holes.** `ArrayLiteral` records elided indices (elements keep the undefined literal so type checking and destructuring are untouched); the interpreter stores `ArrayHole.Instance` and the IL emitters load `$ArrayHole.Instance`, in both simple and spread paths. `1 in a` is now false; map/forEach skip; map preserves the hole.
- **#196 — `Promise.race([])` stays forever pending.** Interp: the never-settling promise was being returned as the task *result*, settling the outer promise with a promise object; now awaits a never-completing TCS. Compiled: the empty-list branch called `SetResult(null)`; now routes a never-completing task through the winning-task await. `all([])`/`allSettled([])`/`any([])` verified correct in both modes.
- **#195 (part 1) — throwing onFulfilled rejects the output promise.** Compiled `$PromiseThen_SM` guarded input `GetResult` and the handler invocation with one try, conflating "input rejected" with "handler threw"; the handler + flatten await now sit in a nested guard try whose catch rejects the builder task (state-1 resume re-enters via first-instruction dispatch, the Roslyn await-inside-try shape). The interpreter had the same conflation for rejecting thenables and faulted the task on handler throws; both fixed. SpeciesConstructor lookup (part 2) is intentionally deferred per the issue, now that `then` is restructured.

## Perf / latency

- **#192 — missing named pipe → immediate ENOENT.** Pre-check existence by enumerating `\\.\pipe\` (safe probe) in both the interpreted socket and the compiled `$NetSocket` emitter; the 5s `Connect` budget now only covers exists-but-busy. `IpcSocket_ConnectErrorCode` drops from ~5-6s to ms.
- **#193 — compiled `fs.watchFile` honors `{ interval }`.** Option extraction handled only `$IHasFields`; compiled object literals are `Dictionary<string, object?>`. Now handles both (mirrors `EmitExtractBooleanOption`). 6.7s → sub-second.
- **#194 — cross-thread timers wake the event loop.** Dropped the `delayMs <= 10` wake heuristic; a blocked `TryTake` could otherwise sleep up to 60s past a new timer''s due time.

## Harness / tests

- **#197 — traced Test262 regens survive worker recycle.** Trace files are now `worker_{id}_{spawnIndex}.log`; the recycled slot no longer reopens a file whose writer is still live.
- **#199 — Pack_WithoutPackageJson tests.** Root cause was environmental, not a regression: `--pack`''s package.json discovery walked to the filesystem root and adopted stray manifests (npm scratch in `%TEMP%`, `{}` in the user profile). The walk now stops at the temp/profile roots (still searched when they are the start directory). All 106 packaging tests pass locally now.

## Validation

- Full suite: **10,517 passed, 0 failed** (previously 10,484 with 2 local packaging failures). Suite wall time dropped 13m18s → 1m6s from the intrinsic-wait fixes.
- Each semantic fix verified against Node behavior with repro scripts in both interpreted and compiled modes.

Noticed in passing (not fixed here): compiled `Promise.any([])` rejects with `name === "Exception"` instead of `AggregateError`.

### DIFF
--- a/Compilation/ExpressionEmitterBase.Operators.cs
+++ b/Compilation/ExpressionEmitterBase.Operators.cs
@@ -164,10 +164,12 @@ public abstract partial class ExpressionEmitterBase
             string name = v.Name.Lexeme;
             double delta = pi.Operator.Type == TokenType.PLUS_PLUS ? 1.0 : -1.0;
 
-            // Load current value, convert to double, add delta, box
+            // Load current value, convert to double, add delta, box.
+            // ConvertToNumber implements ECMA-262 ToNumber (undefined→NaN);
+            // Convert.ToDouble throws InvalidCastException on $Undefined (#190).
             EmitVariable(v);
             EnsureBoxed();
-            IL.Emit(OpCodes.Call, Types.ConvertToDoubleFromObject);
+            IL.Emit(OpCodes.Call, Ctx.Runtime?.ConvertToNumber ?? Types.ConvertToDoubleFromObject);
             IL.Emit(OpCodes.Ldc_R8, delta);
             IL.Emit(OpCodes.Add);
             IL.Emit(OpCodes.Box, typeof(double));
@@ -194,8 +196,8 @@ public abstract partial class ExpressionEmitterBase
             EnsureBoxed();
             IL.Emit(OpCodes.Dup);  // Dup original for expression result
 
-            // Convert to double, add delta, box
-            IL.Emit(OpCodes.Call, Types.ConvertToDoubleFromObject);
+            // Convert to double, add delta, box (ToNumber semantics — see EmitPrefixIncrement)
+            IL.Emit(OpCodes.Call, Ctx.Runtime?.ConvertToNumber ?? Types.ConvertToDoubleFromObject);
             IL.Emit(OpCodes.Ldc_R8, delta);
             IL.Emit(OpCodes.Add);
             IL.Emit(OpCodes.Box, typeof(double));
--- a/Compilation/ExpressionEmitterBase.cs
+++ b/Compilation/ExpressionEmitterBase.cs
@@ -1038,8 +1038,16 @@ public abstract partial class ExpressionEmitterBase : IEmitterContext
             {
                 IL.Emit(OpCodes.Dup);
                 IL.Emit(OpCodes.Ldc_I4, i);
-                EmitExpression(a.Elements[i]);
-                EnsureBoxed();
+                if (a.IsHole(i))
+                {
+                    // Elided position → true ECMA-262 hole, not an undefined element.
+                    IL.Emit(OpCodes.Ldsfld, Ctx.Runtime!.ArrayHoleInstance);
+                }
+                else
+                {
+                    EmitExpression(a.Elements[i]);
+                    EnsureBoxed();
+                }
                 IL.Emit(OpCodes.Stelem_Ref);
             }
 
@@ -1066,8 +1074,15 @@ public abstract partial class ExpressionEmitterBase : IEmitterContext
                     IL.Emit(OpCodes.Newarr, typeof(object));
                     IL.Emit(OpCodes.Dup);
                     IL.Emit(OpCodes.Ldc_I4_0);
-                    EmitExpression(a.Elements[i]);
-                    EnsureBoxed();
+                    if (a.IsHole(i))
+                    {
+                        IL.Emit(OpCodes.Ldsfld, Ctx.Runtime!.ArrayHoleInstance);
+                    }
+                    else
+                    {
+                        EmitExpression(a.Elements[i]);
+                        EnsureBoxed();
+                    }
                     IL.Emit(OpCodes.Stelem_Ref);
                     IL.Emit(OpCodes.Call, Ctx.Runtime!.CreateArray);
                 }
--- a/Compilation/ILEmitter.Properties.Literals.cs
+++ b/Compilation/ILEmitter.Properties.Literals.cs
@@ -34,8 +34,16 @@ public partial class ILEmitter
             {
                 IL.Emit(OpCodes.Dup);
                 IL.Emit(OpCodes.Ldc_I4, i);
-                EmitExpression(a.Elements[i]);
-                EmitBoxIfNeeded(a.Elements[i]);
+                if (a.IsHole(i))
+                {
+                    // Elided position → true ECMA-262 hole, not an undefined element.
+                    IL.Emit(OpCodes.Ldsfld, _ctx.Runtime!.ArrayHoleInstance);
+                }
+                else
+                {
+                    EmitExpression(a.Elements[i]);
+                    EmitBoxIfNeeded(a.Elements[i]);
+                }
                 IL.Emit(OpCodes.Stelem_Ref);
             }
 
@@ -66,8 +74,15 @@ public partial class ILEmitter
                     IL.Emit(OpCodes.Newarr, _ctx.Types.Object);
                     IL.Emit(OpCodes.Dup);
                     IL.Emit(OpCodes.Ldc_I4, 0);
-                    EmitExpression(a.Elements[i]);
-                    EmitBoxIfNeeded(a.Elements[i]);
+                    if (a.IsHole(i))
+                    {
+                        IL.Emit(OpCodes.Ldsfld, _ctx.Runtime!.ArrayHoleInstance);
+                    }
+                    else
+                    {
+                        EmitExpression(a.Elements[i]);
+                        EmitBoxIfNeeded(a.Elements[i]);
+                    }
                     IL.Emit(OpCodes.Stelem_Ref);
                     IL.Emit(OpCodes.Call, _ctx.Runtime!.CreateArray);
                 }
--- a/Compilation/RuntimeEmitter.CoreUtilities.cs
+++ b/Compilation/RuntimeEmitter.CoreUtilities.cs
@@ -3026,6 +3026,18 @@ public partial class RuntimeEmitter
         il.Emit(OpCodes.Isinst, _types.String);
         il.Emit(OpCodes.Brtrue, stringLabel);
 
+        // .NET interop: boxed non-double numerics (float, int, long, decimal, …)
+        // arrive via @DotNetType calls. Convert them numerically rather than
+        // falling through to NaN (EnsureDouble routes through this method).
+        var notConvertibleLabel = il.DefineLabel();
+        il.Emit(OpCodes.Ldloc, argLocal);
+        il.Emit(OpCodes.Isinst, typeof(IConvertible));
+        il.Emit(OpCodes.Brfalse, notConvertibleLabel);
+        il.Emit(OpCodes.Ldloc, argLocal);
+        il.Emit(OpCodes.Call, _types.GetMethod(_types.Convert, "ToDouble", _types.Object));
+        il.Emit(OpCodes.Ret);
+        il.MarkLabel(notConvertibleLabel);
+
         // everything else => NaN
         il.Emit(OpCodes.Br, nanLabel);
 
--- a/Compilation/RuntimeEmitter.FsWatcher.cs
+++ b/Compilation/RuntimeEmitter.FsWatcher.cs
@@ -385,28 +385,47 @@ public partial class RuntimeEmitter
         il.Emit(OpCodes.Ldc_I4, 5007);
         il.Emit(OpCodes.Stloc, intervalLocal);
 
-        // Check if arg1 is options dict with "interval"
+        // Check if arg1 is an options bag with "interval". Compiled object literals
+        // are Dictionary<string, object?> — NOT $IHasFields (only class instances
+        // implement that) — so both shapes must be handled or literal options bags
+        // (the normal calling convention) are silently ignored.
         var afterOptionsLabel = il.DefineLabel();
+        var haveValueLabel = il.DefineLabel();
+        var notHasFieldsLabel = il.DefineLabel();
+        var intervalObjLocal = il.DeclareLocal(_types.Object);
+
+        // if (arg1 is $IHasFields hf) intervalObj = hf.GetProperty("interval")
         il.Emit(OpCodes.Ldarg_1);
         il.Emit(OpCodes.Isinst, runtime.IHasFieldsInterface);
-        il.Emit(OpCodes.Brfalse, afterOptionsLabel);
-
+        il.Emit(OpCodes.Brfalse, notHasFieldsLabel);
         il.Emit(OpCodes.Ldarg_1);
         il.Emit(OpCodes.Castclass, runtime.IHasFieldsInterface);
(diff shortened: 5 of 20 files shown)


nickna:
Follow-ups filed for the two gaps noted in the description: #220 (compiled any([]) rejects with raw Exception instead of AggregateError) and #221 (SpeciesConstructor/NewPromiseCapability in then, split from #195 part 2 — the then restructure in this PR is its prerequisite).

Issue #226, opened 2026-06-10
TS Conformance: remaining Fail-bucket clusters (post-#188/#202 survey)


nickna:
Part of #80. Survey of the 21 Fail + 4 ParseError tests remaining after PRs #201/#203 and the declare-function/extends fixes (PR pending). Clusters by root cause:

## Fixed in the pending PR (tests improved, not yet flipped)
- ~~`declare function` never registered~~ (TS2304 extras on 4+ tests)
- ~~`interface B extends A<Base>`~~ rejected with TS2312
- ~~`interface I extends SomeClass`~~ rejected with TS2312

## Remaining clusters
1. **Overload resolution subtype pass** — `anyAssignabilityInInheritance` (16 extras): `foo(a)` with `a: any` must pick the `(x: any)` overload (tsc's subtype-matching pass runs before the assignability pass; `any` is not a *subtype* of `number`). We pick the first assignable overload → wrong return types cascade into TS2322s.
2. **Parameter defaults referencing earlier parameters** — `genericCallWithObjectTypeArgsAndInitializers`: `function foo4<T, U extends T>(x: T, y: U = x)` → TS2304 on `x`; earlier params must be in scope for later defaults (then the expected TS2322s emerge naturally).
3. **Self-referential `typeof`/initializers** — `assignmentCompatWithObjectMembers`: `var a: { foo: typeof a }` and `var a2 = { foo: a2 }` → TS2304.
4. **Member accessibility in structural compat** — `assignmentCompatWithObjectMembersAccessibility` (16 missing TS2322): private/protected members must make otherwise-identical types incompatible (distinct declarations) in the module-scoped cases.
5. **Index-signature assignability against generic instantiations in generic context** — `assignmentCompatWith{Numeric,String}Indexer2` remainders (`A<T>` vs `{ [x: number]: Derived }` inside `function foo<T extends Base>()`).
6. **Enum assignability details** — `enumAssignability` (6 missing: enum → number-literal-typed positions?), `enumAssignabilityInInheritance` (extra TS2322@L81, missing TS2403 var-redecl checks).
7. **Conditional-type evaluation** — `conditionalTypes1/2`, `conditionalTypesExcessProperties`, `inferTypesWithExtends2` — deep checker work on conditional/infer evaluation.
[...]

PR #227, opened 2026-06-10
feat: ambient functions, interface extends class/generic instantiation, two-pass overload resolution (#226)

nickna:
Part of #80; first installment of the #226 cluster survey.

## What
1. **`declare function` registration** — ambient declarations (single and overloaded) parsed into the pending-overload map and waited for an implementation that never comes, so the name was never defined: every call site produced TS2304. Bodyless ambient declarations now define immediately, re-defining with the grown overload set as consecutive declarations accumulate (`Stmt.Function.IsDeclare` threaded from the parser). Ambients still have no runtime binding — correct TS semantics.
2. **`interface B extends A<Base>`** — an `InstantiatedGeneric` base threw TS2312 and aborted the interface; it's now flattened into a concrete base interface by substituting the type arguments into members/index signatures.
3. **`interface I extends SomeClass`** — legal TypeScript; the class's instance shape (own + inherited fields/methods/getters) becomes the base. A class name in type position resolves to its `Instance` type, so that's unwrapped first.

## Result
No baseline flips yet — each improved test has residual deltas from *other* root causes (all catalogued in #226) — but the diffs shrank across ~8 tests with **0 regressions**:
- `enumAssignabilityInInheritance`: 19 wrong lines → 1
- `assignmentCompatWithObjectMembersAccessibility`: 21 → 16
- `conditionalTypes2`: 14 → 10
- `anyAssignabilityInInheritance`: TS2304 noise replaced by the real remaining issue (overload subtype pass, #226 cluster 1)
- TS2312@L1 extras eliminated from `assignmentCompatWith{Numeric,String}Indexer2`

Full unit suite: **10,525 / 10,525 green**. 7 new unit tests.

### DIFF
--- a/Parsing/AST.cs
+++ b/Parsing/AST.cs
@@ -353,7 +353,7 @@ public abstract record Stmt
     /// IsGenerator indicates this is a generator function (function*) that can yield values.
     /// Decorators contains any @decorator annotations applied to this function/method.
     /// </summary>
-    public record Function(Token Name, List<TypeParam>? TypeParams, string? ThisType, List<Parameter> Parameters, List<Stmt>? Body, string? ReturnType, bool IsStatic = false, AccessModifier Access = AccessModifier.Public, bool IsAbstract = false, bool IsOverride = false, bool IsAsync = false, bool IsGenerator = false, List<Decorator>? Decorators = null, bool IsPrivate = false) : Stmt;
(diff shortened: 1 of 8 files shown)


nickna:
Pushed a second commit (`4ad8c729`): **two-pass overload resolution** — tsc's subtype-matching pass before the assignability pass, so an `any` argument picks the `(x: any)` overload rather than the first assignable one.

That flips `anyAssignabilityInInheritance` **Fail → Pass** (subset Pass 53 → 54, 0 regressions; baseline regenerated). Full unit suite green at 10,532. The PR now covers #226 clusters: declare-function registration, extends class/generic-instantiation, and the overload subtype pass.

Issue #246, opened 2026-06-11
instanceof AbortSignal is unreliable: interpreter false-negative, compiled matches ANY plain object


nickna:
Found while implementing #224 (value-position AbortSignal).

## Symptom

```ts
const s: any = AbortSignal.abort('x');
console.log(s instanceof (AbortSignal as any));   // interp: false (Node: true)
console.log(({} as any) instanceof (AbortSignal as any)); // compiled: TRUE (Node: false)
```

- **Interpreter**: `false` for a real signal — the signal is a `SharpTSAbortSignal` while the `AbortSignal` global is a `SharpTSBuiltInConstructor` sentinel with no brand linkage, so the instanceof walk finds nothing.
- **Compiled**: `true` for ANY plain object — both the `AbortSignal` namespace singleton (post-#224) and abort-signal instances are `Dictionary<string, object>`, and `InstanceOf`'s `Type.IsAssignableFrom` fallback happily matches every dict, including `{}`. A false positive is worse than the old ReferenceError.

## Direction

- Compiled: `InstanceOf` needs to special-case the namespace-singleton dicts (AbortSignal/Intl) — either brand signal dicts (e.g. the existing `_reasonSet` slot) and check that, or refuse `IsAssignableFrom` shortcuts when the constructor operand is a plain dict.
- Interpreter: give `SharpTSBuiltInConstructor` an instanceof hook so the AbortSignal sentinel can brand-check `SharpTSAbortSignal` instances (same shape RegExp/Error already solve via dedicated constructor objects).

Related: #232 (combinator AggregateError instanceof), #233 (built-in subclass branding).

Issue #234, opened 2026-06-10
Symbol is not a first-class global value: typeof Symbol is 'undefined', aliasing fails


nickna:
## Symptom

`Symbol(...)` calls and well-known-symbol member access work, but `Symbol` is not actually bound as a value-position global — it only works as the special-cased head of a call or member expression:

```ts
console.log(typeof Symbol);        // "undefined"  (Node: "function")
console.log(typeof Symbol("x"));   // "symbol"     (works)
console.log(typeof Symbol.iterator); // "symbol"   (works)
console.log(typeof Symbol.species);  // "symbol"   (works)
const f = Symbol;                  // Runtime Error: Undefined variable 'Symbol'.
(Symbol as any).species;           // Runtime Error: Undefined variable 'Symbol'.
```

Same bug class as #208 (AbortSignal/Intl/ReadableStream namespaces not bound as global identifiers), which PR #230 fixed for those namespaces — `Symbol` needs the same treatment: a real global binding whose member access and callability go through the normal paths, so aliasing, `typeof`, and cast expressions behave.

Interpreter mode verified; compiled mode likely shares the gap (same special-casing pattern). #224 already tracks the compiled-mode equivalent for the #208 namespaces; whoever fixes this should check both modes.

Found while scoping #221 (species lookup reads `constructor[Symbol.species]`).

Issue #233, opened 2026-06-10
Guest classes cannot extend built-in constructors (Promise, Array): 'Superclass must be a class'


nickna:
## Symptom

Only `Error` is supported as a built-in superclass. Extending other built-ins fails:

```ts
class MyPromise extends Promise {}        // Runtime Error: Superclass must be a class.
class MyPromise<T> extends Promise<T> {}  // Type Error: Cannot use type arguments with non-generic class 'Promise'
class MyArray extends Array {}            // Runtime Error: Superclass must be a class.
class MyError extends Error {}            // works (special-cased)
```

Two distinct layers fail:

1. **Type checker**: `Promise` in an `extends` clause is not modeled as a generic class — `extends Promise<T>` is rejected with "Cannot use type arguments with non-generic class". (Note `Promise<T>` works fine in type annotations, so this is specific to the heritage-clause path.)
2. **Runtime/interpreter**: the superclass expression must evaluate to a guest `SharpTSClass`; built-in constructor values (Promise, Array) are not classes, so `extends` throws "Superclass must be a class". Compiled mode has the same constraint.

## Why it matters

- This is the hard prerequisite for #221 (Promise SpeciesConstructor / NewPromiseCapability): species resolution is only observable if a guest class can subclass Promise. Without it, SpeciesConstructor always resolves to %Promise% and the machinery is dead code.
- `class X extends Array` is a common JS idiom and a recurring Test262 surface.
- The `Error` special case shows the intended pattern; Promise/Array (and Map/Set once they matter) need the same bridge, including static-side inheritance (`MyPromise.resolve`, `MyArray.from`) and instance branding (`instanceof` both ways).

## Scope note

For Promise specifically, the compiled side is part of the #221 architectural lift (compiled promises are raw `Task<object?>` with no object identity). An interpreter-only Promise bridge without the compiled story would introduce an interp/compiled divergence — the bug class the #205-#216 batch was full of — so this should be planned as one piece.

Found while scoping #221.

Issue #221, opened 2026-06-10
Promise.prototype.then: SpeciesConstructor / NewPromiseCapability not implemented (split from #195)


nickna:
## Background

Split out of #195 part 2, as that issue suggested. Part 1 (the `then` dispatch conflation) was fixed in PR #204, which restructured `$PromiseThen_SM` so the onFulfilled invocation is guarded separately from the input-promise await — the prerequisite for this work.

## Gap

Per ECMA-262 §27.2.5.4, `then` must build its result capability via `SpeciesConstructor(promise, %Promise%)`:

1. Read `promise.constructor` (a poisoned getter must make `then` throw **synchronously** — `test/built-ins/Promise/prototype/then/ctor-poisoned.js`, baselined Fail since PR #187).
2. Read its `Symbol.species` property.
3. Construct the result promise through that constructor via NewPromiseCapability (call it with an executor, capture resolve/reject), so `Promise` subclasses get subclass-typed results from `then`/`catch`/`finally` and the combinators (`all`/`race`/`allSettled`/`any` use NewPromiseCapability too).

SharpTS skips this lookup entirely in both modes.

## Why this is an architectural lift, not a patch

- **Compiled promises have no object identity**: they are raw `Task<object?>` values end-to-end. There is no guest-visible promise object carrying a `constructor` property and no `Symbol.species` lookup point. Supporting this means either wrapping tasks in a promise-object representation or threading constructor metadata alongside every task — touching `then`/`catch`/`finally` plus all four combinators.
- The interpreter''s `SharpTSPromise` would need the same `constructor`/species plumbing plus executor-style NewPromiseCapability.

## Priority

Low. Real-world TypeScript that subclasses `Promise` and depends on species resolution is rare; the affected Test262 tests are already baselined Fail (no regression risk from waiting); and the Test262/JS-conformance track is currently deferred in favor of TS conformance. Park until that track is picked back up.

nickna:
Scoped this during the #213/#220 bugfix batch (2026-06-10). Findings that sharpen the "architectural lift" assessment:

**The machinery is currently unobservable from guest code.** Species resolution only does something when `promise.constructor` can differ from %Promise%, and today it cannot:

- `class MyPromise extends Promise` fails at runtime ("Superclass must be a class"); `extends Promise<T>` is rejected by the checker ("non-generic class"). Only `Error` is special-cased as an extendable built-in → filed #233 as the hard prerequisite.
[...]

PR #247, opened 2026-06-11
Bugfix batch: #222 #223 #224 #228 (+ #238), promise.constructor increment for #221

### LINKED ISSUE #222: Compiled mode: MessageChannel port messaging silently drops messages
**Found while fixing #209 (interpreter-mode MessagePort dispatch).**

In compiled mode, `MessageChannel` port messaging never delivers: a `'message'` listener registered with `port.on(...)` never fires, and the program exits silently.

## Repro (compiled mode)
```ts
const channel: any = new MessageChannel();
channel.port2.on('message', (value: any) => {
    console.log('received: ' + value);   // never prints
    channel.port1.close();
    channel.port2.close();
});
channel.port1.postMessage('hello');
```
`dotnet run -- --compile repro.ts -o repro.dll && dotnet repro.dll` → no output, exit 0.

The interpreter-mode equivalent works after #209 (listener implicitly starts the port and receives the cloned value directly, per Node worker_threads semantics). The compiled $MessageChannel/$MessagePort implementation needs the same semantics:
- `on('message', ...)` implicitly starts the port
- queued messages delivered once started
- listener receives the cloned value directly (not a {data} wrapper)

A regression test exists but is currently interpreter-only: `WorkerThreadsTests.MessageChannel_PortOnMessage_ReceivesPostedValue` — flip it to `ExecutionModes.All` when fixing this.

### LINKED ISSUE #223: Interpreter: web-streams reader promises are raw Tasks — .then()/.catch() property access throws
**Found while fixing #210.**

`SharpTSReadableStreamDefaultReader.GetMember` returns raw `Task<object?>` values for `read()`, `cancel()`, and the `closed` property. `await` handles Tasks fine, but any property access on the returned value throws:

```ts
import { ReadableStream as RS } from 'stream/web';
const rs = new (RS as any)({ start(c: any) { c.enqueue('z'); c.close(); } });
const reader = rs.getReader();
reader.read().then((r: any) => console.log(r));
// Runtime Error: Only instances and objects have properties.
```

## Expected
`reader.read()` returns a Promise object usable with `.then()`/`.catch()`/`.finally()` like any other promise.

## Notes
- `await reader.read()` works (the await path unwraps Tasks), so the gap is specifically property access on a raw Task value.
- Options: wrap the Tasks in `SharpTSPromise` at the reader surface, or teach the interpreter's property-access fallback to treat `Task` like a promise. The former is more localized.
- Same shape likely affects `writer.write()`/`writer.close()`/`writer.ready` on WritableStream writers — audit while fixing.

### LINKED ISSUE #224: Compiled mode: AbortSignal/Intl/web-streams globals unusable in value position
**Found while fixing #208 (interpreter-mode global bindings).**

The compiled path has its own gaps for the same global identifiers:

```ts
const s: any = AbortSignal.abort('r');
console.log(s.aborted, s.reason);   // prints: undefined undefined
const I: any = Intl;                 // ReferenceError: Undefined variable 'Intl'
```

- `AbortSignal.abort(...)` compiles but the returned signal's `aborted`/`reason` read as undefined.
- Bare `Intl` in value position throws `ReferenceError: Undefined variable 'Intl'` at runtime ($Runtime.ThrowUndefinedVariable).
- `ReadableStream`/`WritableStream`/`TransformStream`/`MessageChannel` value-position references and `instanceof` should be audited on the compiled path as well.

Interpreter-mode equivalents are covered by `NamespaceGlobalBindingTests` (interpreter-only) — extend those to `ExecutionModes.All` as compiled support lands.

### LINKED ISSUE #228: Interpreter: exceptions in async callbacks invoked by built-ins are silently swallowed (no unhandled-rejection reporting)
**Found while diagnosing #207.**

When C# built-in code invokes a guest async callback via `callback.Call(interpreter, ...)` (e.g. `SharpTSHttpServer.Listen`, timer callbacks, fs callbacks), the returned promise/Task is discarded. If the async callback throws — synchronously after a resume or via rejection — the exception disappears completely: no output, no crash, and any event-loop Refs the callback would have released stay leaked, which can hang the process at exit.

This is exactly how #207 stayed hidden: the resumed continuation threw `Undefined variable 'server'`, the rejection was dropped, and the symptom was a silent hang.

## Expected
Node semantics: an unhandled promise rejection is reported (`unhandledRejection` event / process crash with a stack trace by default).

## Suggested direction
- Track the Task returned by `ISharpTSAsyncCallable.Call`/`CallAsync` at the built-in invocation sites (or centrally in CallableInterop) and attach a continuation that surfaces faulted tasks: write the error to stderr and set a nonzero exit code, or route through a process-level `unhandledRejection` emitter if/when one exists.
- An audit of `callback.Call` sites in Runtime/Types and Runtime/BuiltIns that can receive async callables would show the blast radius.

### LINKED ISSUE #238: Interpreter: top-level promise continuations escape the event loop (sync context installed only in RunEventLoop)
## Symptom

A `.then()` chain started at module top level can have its continuation resume on a **thread-pool thread** instead of the interpreter's event loop. Because the continuation then mutates the interpreter's ambient environment concurrently with the main thread, the symptom is a spurious `Runtime Error: Undefined variable 'r'` (the then-callback's own parameter!) — or silently missing output, depending on the race.

```ts
import { ReadableStream as RS } from 'stream/web';
const rs = new (RS as any)({ start(c: any) { c.enqueue('z'); c.close(); } });
const reader = rs.getReader();
reader.closed.then(() => console.log('closed-resolved'));
reader.read().then((r: any) => console.log('read:', r.value, r.done));
```

In the test harness this fails deterministically (`Undefined variable 'r'`); via the CLI it happens to win the race and print both lines.

## Root cause

`InterpreterSynchronizationContext` was installed **only inside `RunEventLoop`** (Execution/Interpreter.cs), which runs *after* all module top-level statements. Any async built-in (e.g. `PromiseBuiltIns.ThenImpl`) that suspends during top-level execution captures `SynchronizationContext.Current == null`, so its resume is posted to the thread pool. Two code comments (RunEventLoop and SharpTSReadableStream.ReadInternal) claimed `InterpretModules` installs the context 'at the top' — it didn't; the comment was stale.

This is the same ambient-environment-across-threads bug class as #207/#228.

## Fix

Install the sync context at the start of `Interpret` and `InterpretModules` (restored in `finally`), so continuations post to the interpreter callback queue from the very first statement. `WaitForPromise`/`TickEventLoop` already drain that queue on the main thread.

Found while adding tests for #223.

nickna:
One commit per issue; full suite green after each (10,596 tests at HEAD).

### #238 — interpreter: top-level promise continuations escaped the event loop *(found & filed during this batch)*
`InterpreterSynchronizationContext` was installed only inside `RunEventLoop`, which runs **after** module top-level statements. Async built-ins suspending at top level (e.g. `ThenImpl`) captured a null context, resumed on thread-pool threads, and raced the main thread's ambient environment — spurious `Undefined variable` for a then-callback's own parameter, or silently dropped output. The context is now installed at the start of `Interpret`/`InterpretModules` (restored in `finally`). Two stale comments claiming this already happened are corrected.

### #223 — interpreter: web-streams surfaces returned raw Tasks
`reader.read()/cancel()/closed`, `stream.cancel()/pipeTo()`, writable `abort()/close()`, writer `write()/close()/abort()/ready/closed` all returned raw `Task<object?>` — awaitable, but any `.then()/.catch()` access threw *"Only instances and objects have properties"*. All wrapped in `SharpTSPromise` at the member-dispatch boundary; `closed`/`ready` cache their promise object.

### #228 — interpreter: async-callback rejections silently swallowed
New `Interpreter.ObserveDiscardedCallbackResult`/`InvokeGuestCallback`: built-in invocation sites that discard a guest callback's result (timers, fs/dns/crypto completion callbacks, EventEmitter listeners, http server callbacks) now observe the promise and report a faulted one to stderr (`Unhandled promise rejection: …` + rejection stack), setting `HadUnhandledRejection`; the CLI exits 1 after the loop drains — Node's default behavior. The #207 silent-hang shape (`throw` after `await` in a timer callback) now reports loudly.

### #224 — compiled: AbortSignal/Intl/web-streams globals in value position
- `AbortSignal.abort('r')` signals: `aborted`/`reason`/`onabort` now resolve through dynamically-typed receivers (GetProperty dict-miss dispatch keyed on the `_reasonSet` internal slot).
- Bare `AbortSignal`/`Intl` resolve to lazily-populated namespace singletons (Math/JSON pattern) wrapping the existing helpers.
- Bare `ReadableStream`/`WritableStream`/`TransformStream` emit their pure-IL Type tokens (typeof/alias/instanceof/construction work).
- `new` on dynamic values: new `ConstructDynamicValue` (Type → Activator, callable → NewOnFunction, plain object → TypeError) replaces the silent-null fallback in state-machine emitters; `NewOnFunction` throws TypeError for plain-object callees; `IsConstructor` exempts the `CreateIntl*` factories (constructors per ECMA-402), so `const I = Intl; new I.NumberFormat('en-US')` works.
- `NamespaceGlobalBindingTests` flip to `ExecutionModes.All`.

### #222 — compiled: MessageChannel port messaging delivered nothing
Replaced the Dictionary-with-`new object()`-placeholder channel with real pure-IL `$MessagePort`/`$MessageChannel` types modeled on `$BroadcastChannel`: `$MessagePort` extends the emitted `$EventEmitter` and overrides its `OnListenerAdded` hook for implicit-start-on-'message'; `postMessage` structured-clones to the partner's queue with async $EventLoop delivery (listener receives the value directly, queued messages delivered in order after start); started ports Ref the loop, `close()` Unrefs. Value-position `MessageChannel`/`MessagePort` + instanceof now work too. Tests flip to `ExecutionModes.All` + new queued-before-listener ordering test.

### #221 — Promise SpeciesConstructor/NewPromiseCapability
Implemented the observable subset available today: `promise.constructor === Promise` holds by identity in **both** modes (ECMA-262 §27.2.5.1). Full species dispatch is unobservable until its prerequisites land — #233 (guest classes cannot extend Promise, so SpeciesConstructor always resolves to %Promise%) and #234 (Symbol not first-class) — per the analysis already recorded on those issues. then/catch/finally continue constructing through %Promise%.

## New issues filed during this batch
- #238 — sync-context gap (fixed here)
- #246 — `instanceof AbortSignal` unreliable: interpreter false-negative, compiled matches ANY plain object (pre-existing, surfaced while testing #224)

Fixes #222. Fixes #223. Fixes #224. Fixes #228. Fixes #238.

### DIFF
--- a/Compilation/EmittedRuntime.cs
+++ b/Compilation/EmittedRuntime.cs
@@ -584,6 +584,7 @@ public class EmittedRuntime
     // Invocation methods
     public MethodBuilder InvokeValue { get; set; } = null!;
     public MethodBuilder InvokeMethodValue { get; set; } = null!;
+    public MethodBuilder ConstructDynamicValue { get; set; } = null!;
     public MethodBuilder GetSuperMethod { get; set; } = null!;
 
     // Dynamic JS iterator-protocol bridge (.next()/.return()) for any-typed
@@ -2098,6 +2099,13 @@ public class EmittedRuntime
     // $AsyncLocalStorage support
     public Type TSAsyncLocalStorageType { get; set; } = null!;
     public ConstructorBuilder TSAsyncLocalStorageCtor { get; set; } = null!;
+
+    // Value-position namespace singletons (#224). Null when the matching
+    // feature flag is off — EmitVariable guards on the populate method.
+    public FieldBuilder? AbortSignalNamespaceField { get; set; }
+    public MethodBuilder? AbortSignalNamespacePopulate { get; set; }
+    public FieldBuilder? IntlNamespaceField { get; set; }
+    public MethodBuilder? IntlNamespacePopulate { get; set; }
 
     // $AbortController / $AbortSignal support
     public MethodBuilder FireAbortEvent { get; set; } = null!;
--- a/Compilation/ExpressionEmitterBase.cs
+++ b/Compilation/ExpressionEmitterBase.cs
@@ -1348,44 +1348,42 @@ public abstract partial class ExpressionEmitterBase : IEmitterContext
         }
         else
         {
-            // Fallback: try to load via resolver
-            var stackType = Resolver.TryLoadVariable(className);
-            if (stackType != null)
+            // Dynamic-callee fallback (#224): the callee is not a known class
+            // name, so evaluate it as a VALUE (aliased built-in constructor,
+            // namespace-singleton member like `I.NumberFormat`, arbitrary
+            // expression) and construct through ConstructDynamicValue, which
+            // dispatches Type → Activator, callable → InvokeValue, and throws
+            // TypeError for non-constructables. Replaces the old behavior of
+            // silently yielding null when the name didn't resolve.
+            EmitExpression(n.Callee);
+            EnsureBoxed();
+            var ctorTemp = IL.DeclareLocal(typeof(object));
+            IL.Emit(OpCodes.Stloc, ctorTemp);
+
+            List<LocalBuilder> argTemps = [];
+            foreach (var arg in n.Arguments)
             {
-                var typeTemp = IL.DeclareLocal(typeof(object));
-                IL.Emit(OpCodes.Stloc, typeTemp);
-
-                List<LocalBuilder> argTemps = [];
-                foreach (var arg in n.Arguments)
-                {
-                    EmitExpression(arg);
-                    EnsureBoxed();
-                    var temp = IL.DeclareLocal(typeof(object));
-                    IL.Emit(OpCodes.Stloc, temp);
-                    argTemps.Add(temp);
-                }
-
-                IL.Emit(OpCodes.Ldloc, typeTemp);
-                IL.Emit(OpCodes.Ldc_I4, n.Arguments.Count);
-                IL.Emit(OpCodes.Newarr, Ctx.Types.Object);
-
-                for (int i = 0; i < argTemps.Count; i++)
-                {
-                    IL.Emit(OpCodes.Dup);
-                    IL.Emit(OpCodes.Ldc_I4, i);
-                    IL.Emit(OpCodes.Ldloc, argTemps[i]);
-                    IL.Emit(OpCodes.Stelem_Ref);
-                }
-
-                var createInstanceMethod = Ctx.Types.GetMethod(Ctx.Types.Activator, "CreateInstance", Ctx.Types.Type, Ctx.Types.ObjectArray);
-                IL.Emit(OpCodes.Call, createInstanceMethod!);
-                SetStackUnknown();
+                EmitExpression(arg);
+                EnsureBoxed();
+                var temp = IL.DeclareLocal(typeof(object));
+                IL.Emit(OpCodes.Stloc, temp);
+                argTemps.Add(temp);
             }
-            else
+
+            IL.Emit(OpCodes.Ldloc, ctorTemp);
+            IL.Emit(OpCodes.Ldc_I4, n.Arguments.Count);
+            IL.Emit(OpCodes.Newarr, Ctx.Types.Object);
+
+            for (int i = 0; i < argTemps.Count; i++)
             {
-                IL.Emit(OpCodes.Ldnull);
-                SetStackType(StackType.Null);
+                IL.Emit(OpCodes.Dup);
+                IL.Emit(OpCodes.Ldc_I4, i);
+                IL.Emit(OpCodes.Ldloc, argTemps[i]);
+                IL.Emit(OpCodes.Stelem_Ref);
             }
+
+            IL.Emit(OpCodes.Call, Ctx.Runtime!.ConstructDynamicValue);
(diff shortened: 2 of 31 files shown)


nickna:
Rebased onto current main (post-#240/#241/#243/#244). Conflict resolution notes for review:

- PR #243 moved `TryEmitBuiltInClassType` into `ExpressionEmitterBase` so state-machine emitters share it — my web-streams (#224) and MessageChannel/MessagePort (#222) entries now live in that shared base method instead of an ILEmitter-local copy, so they resolve inside async/generator bodies too.
- The AbortSignal/Intl value-position branches were likewise hoisted into a shared `TryEmitNamespaceSingleton` helper, called from both `ILEmitter.EmitVariable` and the base `TryEmitGlobalVariable` (the #232-style state-machine path).

Full suite green after rebase: 10,675 passed / 0 failed. Re-verified the #222 and #224 compiled repros and the #221 constructor identity manually.

Issue #297, opened 2026-06-11
Compiled mode: dynamic property access does not resolve members inherited from a base class (GetProperty has no base-class delegation)


nickna:
## Summary

In compiled (IL) mode, the per-class `$IHasFields.GetProperty` helper that backs DYNAMIC property access only dispatches a class's **own** members and returns `$Undefined` otherwise — it never delegates to the base class. So a member (method, getter, typed property) **inherited** from a base class resolves to `undefined` under dynamic dispatch, and a call throws `TypeError: undefined is not a function`.

Statically-typed receivers dodge this because `obj.method()` emits a direct virtual `callvirt` that resolves the inherited method via .NET dispatch. But any access that routes through `$Runtime.GetProperty` hits the gap:
- an inherited member on an `any`-typed receiver, and
- **every** method call on a class-EXPRESSION instance (always anonymously typed → always dynamically dispatched).

## Repro

```typescript
class Animal { constructor(public name: string) {} speak(): string { return this.name + " sound"; } }
class Dog extends Animal { constructor(name: string) { super(name); } }
const d: any = new Dog("Fido");
console.log(d.speak());   // compiled: TypeError: undefined is not a function; interp: "Fido sound"
```

```typescript
const Animal = class { constructor(public name: string) {} speak(): string { return this.name + " sound"; } };
const Dog = class extends Animal { constructor(name: string) { super(name); } };
console.log(new Dog("Fido").speak());   // compiled: throws; interp: ok
```

## Fix

At the end of the emitted `GetProperty` (both class-declaration and class-expression paths in `ILCompiler.Classes.HasFields.cs`), when the member isn't found locally and the base is an emitted user class with a `$IHasFields` stub, emit a non-virtual `call base.GetProperty(name)` (resolving the method for constructed-generic bases) instead of returning `$Undefined`. The chain terminates at the topmost emitted class. Fixed in the #287 PR.

## Remaining (follow-up)

This fix covers reads (`GetProperty`). `SetProperty` (inherited setters via dynamic dispatch) and `HasProperty` (`in` for inherited members) still lack base delegation — left as-is here because of the own-property-vs-inherited-setter and per-level `_fields` subtleties. Worth a separate pass.

## Environment
- `origin/main` @ 7eab2cdf; .NET 10.0, Windows 11

nickna:
[...]

Issue #302, opened 2026-06-11
compiled: lodash init fails after #276 — getNative path throws 'object is not a function'


nickna:
## Summary

With #276 fixed (value-form `Math.max`/`JSON.stringify` wrappers now populated), compiled-mode lodash initialization gets one step further but still fails. `RealPackageSmokeTests.Lodash_Compiled` — which only asserts that lodash *loads* (`typeof _ === "function"`) — now throws during module init:

```
System.Exception: TypeError: object is not a function
    at $Runtime.InvokeMethodValue(Object, Object, Object[])
    at InvokeStub_<>c__InnerFuncDC378_getNative.Invoke(...)
    ...
    at $Module_lodash.$Initialize()
    at $Module_lodash.$GetExports()
```

The failure is inside lodash's `getNative(object, key)` helper (`baseIsNative` / `getValue`), invoking something that resolves to a non-callable object in compiled mode — a value-form dispatch gap distinct from #276's Math/JSON singletons.

## Repro

```js
const _ = require('lodash');
console.log(typeof _);     // interp: "function"; compiled: TypeError: object is not a function
```

`Lodash_Interpreter` passes; only the compiled path fails.

## Status

`Lodash_Compiled` remains `[SkippableFact(Skip=...)]`, now pointing at this issue instead of #276. Unskip when the getNative value-form dispatch is resolved. (`_.chunk`/`_.flatten` returning wrong values in compiled mode — noted in the test — is a further, separate behavioral gap.)

## Environment
- Branch builds on `origin/main` @ 0f409b43 + the #276/#281/#296/#297/#300 batch; .NET 10.0, Windows 11

nickna:
Closing in favor of #307. Triage during PR #308 disproved this issue's hypothesis (a value-form dispatch gap in lodash's `getNative`). The real root cause is a compiler closure bug — inner function *declarations* nested inside an arrow / function expression capture enclosing locals by value-snapshot at hoist time instead of by live reference, so `getValue` / `Object` / `symToStringTag` read back `null` (`typeof null === "object"` → "object is not a function"). #307 has the minimal repro and a concrete fix plan (a live `$arrowScopeDC` reference for inner functions, mirroring `$functionDC`). `Lodash_Compiled` stays skipped, pointing at #307.

Issue #300, opened 2026-06-11
Compiled mode: class accessor body cannot resolve a captured top-level variable (ReferenceError)


nickna:
## Summary

In compiled (IL) mode, a class **accessor** body (getter or setter) that references a top-level binding — a captured `let`/`const`, a same-module ESM export, or an import — throws `ReferenceError: Undefined variable` at runtime. Ordinary **method** bodies in the same class resolve the identical reference correctly.

## Repro

```typescript
let counter = 5;
class D { get tag(): string { return "v" + counter; } }
console.log(new D().tag);   // interp: "v5"; compiled: ReferenceError: Undefined variable 'counter'
```

A normal method dodges it:

```typescript
let counter = 5;
class A { m(): string { return "v" + counter; } }   // compiled: "v5" — fine
```

Also reproduces for symbol-keyed accessors (the body, not the key — both well-known and module-local keys), which is how it surfaced.

## Root cause

`ILCompiler.EmitAccessorBody` (`Compilation/ILCompiler.Classes.Accessors.cs`) builds its `CompilationContext` without the four top-level-variable-access properties that **every other** body-emission context sets — `TopLevelStaticVars`, `CapturedTopLevelVars`, `EntryPointDisplayClassFields`, `EntryPointDisplayClassStaticField`. Without them, a bare top-level identifier falls through to the dynamic runtime lookup and `$Runtime.ThrowUndefinedVariable` fires. The accessor path was the lone omission among methods, constructors, static blocks, functions, arrows, async, and generators.

## Fix

Set the four properties on the accessor context (using `BuildClassMethodTopLevelStaticVarsForModule`, matching the instance-method emit path so same-module ESM exports also resolve).

## Environment
- `origin/main` @ 0f409b43; .NET 10.0, Windows 11
- Found while verifying #282 / #281 (symbol-keyed accessors).

nickna:
Scope expands slightly: the same omission exists in `CreateClassExpressionContext` (all class-expression member bodies — methods and accessors), not just `EmitAccessorBody`. Both are fixed together in the linked PR.

nickna:
Fixed in #303 (merged): EmitAccessorBody and CreateClassExpressionContext now wire the four top-level-variable-access properties every other body-emission context sets.

Issue #296, opened 2026-06-11
Compiled mode: class expression extending another class expression never sets the parent type (ILVerify CallCtor/ThisUninitReturn, wrong instanceof)


nickna:
## Summary

In compiled (IL) mode, a class expression that extends **another class expression** (`const A = class {}; const B = class extends A {}`) silently extends `System.Object` instead of `A`. The base-type resolver matches the source variable name (`"A"`) against `_classExprs.Names`, which stores the *generated* names (`$ClassExpr_N`), so the lookup never hits and `SetParent` is never called. Meanwhile `super(...)` still chains the real parent ctor, so the verifier sees a base-ctor `call` on a `this` whose declared base is `Object` → `CallCtor` + `ThisUninitReturn` + `StackUnexpected`. `instanceof` against the parent also returns `false`.

## Repro

```typescript
const A = class { constructor(public x: number) {} };
const B = class extends A { constructor() { super(3); } };
console.log(new B().x);            // 3 (runs)
console.log(new B() instanceof A); // should be true, was false
```

```
$ dotnet run -- --compile ce.ts -o ce.dll --verify
IL verification found 3 error(s):
  [IL Error] $ClassExpr_2..ctor: CallCtor ...
  [IL Error] $ClassExpr_2..ctor: StackUnexpected ...
  [IL Error] $ClassExpr_2..ctor: ThisUninitReturn ...
```

When the parent is a class **declaration** (not expression) the lookup falls through to `ClassRegistry` and works, so only class-expr-extends-class-expr is affected.

## Fix

Consult `_classExprs.VarToClassExpr` (variable-name → class expr) in the base-type resolver (`ILCompiler.Classes.ClassExpressions.cs`), and resolve the implicit-base-ctor via the registered `ConstructorBuilder` rather than `Type.GetConstructor` on the unbaked parent `TypeBuilder`.

Found while fixing #287 (same CallCtor family). Fixed in the #287 PR.

## Environment
- `origin/main` @ 7eab2cdf; .NET 10.0, Windows 11

nickna:
Fixed in #303 (merged): class-expr base-type resolver consults VarToClassExpr and resolves the implicit base ctor via the registered ConstructorBuilder; IL verifies and instanceof is correct.

PR #303, opened 2026-06-11
Compiled & interp class/Math fixes: #297 #296 #287 #281 #300 #288 (+ verify #282/#276); supersedes #301

nickna:
Batch of compiled-mode (IL) class fixes. Each issue is a separate commit; full suite green (10908 passed, 1 intentional skip).

### #297 — inherited members under dynamic dispatch (reads)
The per-class `$IHasFields.GetProperty` helper only dispatched a class's **own** members and returned `$Undefined` otherwise, so a member **inherited** from a base class resolved to `undefined` under dynamic dispatch (an `any`-typed receiver, or *any* method call on a class-expression instance — always anonymously typed). `GetProperty` now emits a non-virtual `call base.GetProperty(name)` when the member isn't found locally and the base is an emitted user class, terminating at the topmost emitted class. Reads only; `SetProperty`/`HasProperty` base delegation remain the follow-up the issue notes.

### #296 — class expression extending another class expression
The base-type resolver matched the source variable name against `_classExprs.Names` (which holds the *generated* `$ClassExpr_N` names), so `SetParent` was never called: the child silently extended `System.Object` while `super(...)` chained the real parent ctor → ILVerify `CallCtor`/`ThisUninitReturn`, and `instanceof` returned `false`. Now consults `_classExprs.VarToClassExpr` and resolves the implicit base ctor via the registered `ConstructorBuilder`.

### #287 (prerequisite of the above CallCtor family)
A subclass with an explicit `super(...)` emitted `Object..ctor` before `super()` chained the real parent ctor → ILVerify `CallCtor`. The Object ctor is now skipped when a user-class superclass with an emitted ctor exists (mirroring the Error/Array/Promise handling).

### #281 — symbol-keyed computed accessors on class expressions
Class expressions had no hook for the symbol-accessor machinery added for class declarations in #266, so `const C = class { get [Symbol.toStringTag]() {…} }` threw a `CompileException`. The class-expression emitter now mirrors the declaration path: define synthetic `$sym_get/set_N` methods, emit their bodies in the class-expression context, and register them in the class `.cctor` via the shared `EmitSymbolAccessorRegistrations`. Runtime registry + dispatch from #266 reused unchanged.

### #300 — accessor & class-expression bodies couldn't see top-level variables (new, found here)
`EmitAccessorBody` and `CreateClassExpressionContext` were the only two body-emission contexts that omitted the four top-level-variable-access properties every other context sets, so a getter/setter or class-expression member body referencing a captured `let`/`const`, a same-module export, or an import threw `ReferenceError: Undefined variable` in compiled mode. Both contexts now wire them up.

## Verified already-fixed on main (regression tests added / scope confirmed)

- **#282** (module-local Symbol-keyed accessor): passes on main; existing coverage only exercised well-known symbols, so a both-modes getter+setter regression test for the module-local case is added.
- **#276** (value-form `Math.max`/`JSON.stringify`): passes on main with comprehensive existing coverage. Un-skipping `Lodash_Compiled` revealed lodash init now gets past `context.Math.max` but fails one step deeper in `getNative` — filed as **#302** and the skip re-pointed there.

## Issues filed during this batch
- **#300** — accessor/class-expr bodies can't resolve top-level variables (fixed here).
- **#302** — compiled lodash init fails in `getNative` after #276 (separate value-form dispatch gap).

## Testing
`dotnet build` clean; `dotnet test` → 10908 passed, 1 skipped (`Lodash_Compiled`, gated on #302). New regression tests cover inherited dynamic dispatch (decl + class-expr, multi-level), class-expr-extends-class-expr verification, derived-ctor verification, symbol accessors on class expressions (getter/setter/well-known/module-local/static), and accessor/class-expr top-level-variable resolution.

### DIFF
--- a/Compilation/ILCompiler.Classes.Accessors.cs
+++ b/Compilation/ILCompiler.Classes.Accessors.cs
@@ -203,7 +203,16 @@ public partial class ILCompiler
             ClassExprBuilders = _classExprs.Builders,
             IsStrictMode = _isStrictMode,
             // Registry services
-            ClassRegistry = GetClassRegistry()
+            ClassRegistry = GetClassRegistry(),
+            // Module-level / captured top-level variable access. Without these an
+            // accessor body that references a top-level binding (a captured `let`,
+            // a same-module export, an import) throws ReferenceError at runtime —
+            // every other body-emission context (methods, ctors, functions, …)
+            // sets them; the accessor path was the lone omission.
+            TopLevelStaticVars = BuildClassMethodTopLevelStaticVarsForModule(_modules.CurrentPath),
+            CapturedTopLevelVars = BuildCapturedTopLevelVarsForModule(_modules.CurrentPath),
+            EntryPointDisplayClassFields = BuildEntryPointDisplayClassFieldsForModule(_modules.CurrentPath),
+            EntryPointDisplayClassStaticField = _closures.EntryPointDisplayClassStaticField
         };
 
         // Add class generic type parameters to context
--- a/Compilation/ILCompiler.Classes.ClassExpressions.cs
+++ b/Compilation/ILCompiler.Classes.ClassExpressions.cs
@@ -80,6 +80,18 @@ public partial class ILCompiler
             {
                 // Found in class declarations
             }
+            else if (_classExprs.VarToClassExpr.TryGetValue(superclassName, out var parentClassExpr)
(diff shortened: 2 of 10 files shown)


nickna:
Adds the #288 commit (Math value-form identity + `Object.keys/values/entries(Math)` parity) cherry-picked from #301, so this PR is a strict superset of #301. Per the decision to keep #303 as the superset, **#301 is being closed in favor of this PR.**

Issue #288, opened 2026-06-11
Interpreter: value-form Math singleton diverges from compiled (Object.keys(Math) throws; Math.max !== Math.max)


nickna:
## Summary

After #276 fixed compiled-mode value-form access to the Math/JSON singletons, two interpreter-side divergences remain when `Math` is used as a *value*:

```ts
const m: any = Math;
console.log(m.max(1, 2));          // 2 in both modes  ✓
console.log(m.max === Math.max);   // compiled: true   interp: false
console.log(Object.keys(m).length);// compiled: 0      interp: throws
```

## Details

1. **Method identity** — `Math.max === Math.max` is `true` per ECMA-262 (built-in methods are single objects). Compiled mode returns the same identity-cached `$TSFunction` (post-#276). The interpreter synthesizes a fresh wrapper per read, so `m.max === Math.max` is `false`.

2. **`Object.keys(Math)`** — Math's own properties are all non-enumerable, so `Object.keys(Math)` is `[]`. Compiled mode returns an empty array; the interpreter throws `Runtime Error: Object.keys() requires an object argument` — it doesn't treat the Math singleton as a valid object receiver.

Both are minor and interpreter-only; compiled mode matches the spec. Filed for parity tracking. The compiled-mode regression test `BuiltInSingletonValueFormTests` scopes the identity/non-enumerable assertions to `CompiledOnly` because of these gaps.

## Environment
- `origin/main` @ ef4b0a46 + #276; .NET 10.0, Windows 11
- Found while fixing #276 (2026-06-11 batch).

nickna:
Fixed in #303 (merged via the #301 commit): Math value-form method identity is stable and Object.keys/values/entries(Math) match compiled mode; the two assertions now run in both modes.

Issue #298, opened 2026-06-11
Compiled mode: Object.values(Math)/Object.entries(Math) enumerate non-enumerable built-in methods


nickna:
## Summary

In compiled (IL) mode, `Object.values(Math)` and `Object.entries(Math)` enumerate Math's built-in methods (abs, max, …), which are non-enumerable per ECMA-262 §17. `Object.keys(Math)` correctly returns only user-assigned extras, so keys and values/entries are inconsistent with each other.

## Repro

```typescript
const m: any = Math;
m.foo = 42;
console.log(Object.keys(m));        // [foo]            (correct)
console.log(Object.values(m));      // [[Function]×35, 42]   (wrong — built-ins leak)
console.log(Object.entries(m).length); // 37             (wrong — should be 1)
```

Expected (per spec, and matching the interpreter after #288): `Object.values(Math)` and `Object.entries(Math)` return only the own enumerable (user-assigned) properties.

## Notes

The compiled GetKeys path already filters to enumerable descriptors for Object.keys; the values/entries paths don't apply the same filter to the Math/JSON singleton dictionaries. Found while fixing #288 (interpreter Math value-form parity), where the interpreter's Object.values/entries(Math) were made spec-correct — the corresponding interp assertions are scoped InterpretedOnly until this is fixed.

## Environment
- `origin/main` @ 0f409b43; .NET 10.0, Windows 11

PR #305, opened 2026-06-11
fix(interp): Math value-form identity + Object.keys/values/entries pa…

nickna:
…rity (#288)

Two interpreter-only divergences from compiled mode when Math is used as a value:

1. Method identity — `m.max === Math.max` was false. The namespace path (`Math.max`) returned the unbound BuiltInMethod from MathBuiltIns, while the instance path (`m.max`, m holding Math) returned that method Bound to the singleton receiver, so the two access forms yielded distinct wrappers. The Math namespace's static GetMethod now binds to the SharpTSMath singleton; Bind caches by receiver, so both paths return the same method and identity is stable (ECMA-262: built-in methods are single objects).

2. Object.keys/values/entries(Math) threw "requires an object argument". The Math singleton is now a recognized receiver: its built-in members are non-enumerable per ECMA-262 §17, so the own enumerable properties are exactly the user-assigned extras (empty in the common case) — matching compiled Object.keys(Math).

The two compiled-only assertions in BuiltInSingletonValueFormTests now run in both modes.

### DIFF
--- a/Runtime/BuiltIns/BuiltInRegistry.cs
+++ b/Runtime/BuiltIns/BuiltInRegistry.cs
@@ -333,7 +333,12 @@ public sealed class BuiltInRegistry
             Name: "Math",
             IsSingleton: true,
             SingletonFactory: () => SharpTSMath.Instance,
-            GetMethod: name => MathBuiltIns.GetMember(name) as BuiltInMethod
+            // Bind to the singleton so the namespace path (`Math.max`) returns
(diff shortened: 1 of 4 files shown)


nickna:
Superseded: the #288 fix already landed via #303. Production files are byte-for-byte identical to main, and main's test file is a strict superset (includes the #288 both-mode assertions plus #298/#299 tests this branch predates). Nothing to merge.

Issue #407, opened 2026-06-13
Worker stdin/stdout/stderr/resourceLimits options are parsed but never implemented


nickna:
Observed while fixing #380.

### Gap
`SharpTSWorker` reads `stdin`, `stdout`, `stderr`, and `resourceLimits` from the options bag into private fields (`_stdin`, `_stdout`, `_stderr`, `_resourceLimits`), but **nothing consumes them** in either interpreter or compiled mode — they are inert dead fields. So `new Worker(f, { stdout: true })`, `resourceLimits`, etc. have no observable effect.

Additionally, `_resourceLimits` is typed `SharpTSArray?`, but Node's `resourceLimits` is an object (`{ maxOldGenerationSizeMb, ... }`), not an array — so `as SharpTSArray` would always yield null even in interpreter mode.

### Suggested approach
Either implement the stdio piping / resource-limit honoring, or document them as unsupported and stop reading them. If implementing, fix the `_resourceLimits` type to read the object shape.

Related: #380, #354.

Issue #406, opened 2026-06-13
Compiled Worker: transferList / cross-boundary MessagePort transfer is ignored


nickna:
Follow-up to #380.

### Gap
With #380, the compiled worker options bag (`workerData`/stdio) is now marshaled correctly. `transferList` is **read** generically too, but it stays typed as `SharpTSArray?` in `SharpTSWorker`, and:

- A compiled `transferList` is a `List<object?>` (compiled array), not a `SharpTSArray`, so `ReadOption(...) as SharpTSArray` yields null and the transfer list is dropped.
- Even if it were marshaled, `StructuredClone.Clone(value, SharpTSArray? transfer)` only transfers `SharpTSMessagePort`. Compiled `MessagePort` values are the emitted `$MessagePort` type, not `SharpTSMessagePort`, so cross-boundary MessagePort transfer from a compiled parent is unsupported.

This is intentionally out of scope for #380 (which fixes the headline `workerData === undefined` bug). `workerData` still structured-clones correctly without a transfer list.

### Repro
```ts
// compiled main.ts
import { Worker, MessageChannel } from "worker_threads";
const { port1, port2 } = new MessageChannel();
new Worker(__dirname + "/w.ts", { workerData: { port: port1 }, transferList: [port1] });
```
The transferred port is not delivered to the worker in compiled mode.

### Suggested approach
- Widen `StructuredClone.Clone`'s transfer parameter to accept compiled `List<object?>` and the emitted `$MessagePort` representation.
- Define a cross-runtime MessagePort transfer contract so a compiled `$MessagePort` can be adopted by the worker's interpreter.

Related: #380, #354, #329.

PR #408, opened 2026-06-13
fix(compile): honor compiled Worker workerData option (#380)

### LINKED ISSUE #380: Compiled Worker: options bag (workerData/transferList/stdio) is ignored
Follow-up to #354 (PR #379), which enabled compiled `new Worker(...)`.

### Gap
In compiled mode the worker is constructed via `SharpTSWorker.CreateForCompiledLoop(filename, options, ...)`, which does `options as SharpTSObject`. A compiled object literal (e.g. `new Worker(f, { workerData: 42 })`) is **not** a `SharpTSObject`, so the cast yields null and the entire options bag — `workerData`, `transferList`, `stdin/stdout/stderr`, `resourceLimits` — is silently dropped.

So a compiled worker spawned with `workerData` sees `workerData === undefined`, and there is no transfer-list / stdio support.

By contrast the interpreter path passes a real `SharpTSObject` and honors all of these.

### Repro
```ts
// main.ts (compiled)
import { Worker } from "worker_threads";
new Worker(__dirname + "/w.ts", { workerData: 123 });
// w.ts
import { workerData } from "worker_threads";
console.log(workerData); // prints undefined compiled; 123 interpreted
```
(Note compiled `worker_threads.workerData` also currently returns null unconditionally — see `WorkerThreadsModuleEmitter.EmitWorkerData`.)

### Suggested approach
- Marshal the compiled options object into a form `SharpTSWorker` can read in `CreateForCompiledLoop` (read `workerData`/`transferList`/… via the dynamic `$Runtime.GetProperty` rather than requiring a `SharpTSObject`), and structured-clone `workerData` across the boundary.
- Wire compiled `worker_threads.workerData` to return the cloned value in worker context instead of null.

Related: #354, PR #379, #329.

nickna:
Closes #380.

## Problem
A compiled program that spawns `new Worker(f, { workerData: 123 })` saw `workerData === undefined` inside the worker. The options bag is a compiled object literal (`Dictionary<string, object?>`), but `SharpTSWorker.CreateForCompiledLoop` cast it with `options as SharpTSObject`, which yields null for that type — silently dropping `workerData`, `transferList`, and stdio. The interpreter path passes a real `SharpTSObject`, so it was unaffected.

## Fix
- `SharpTSWorker` ctor now takes `object?` options and reads each property through a new representation-agnostic `ReadOption` helper that handles **both** a `SharpTSObject` (interpreter) and an `IDictionary<string, object?>` (compiled literal).
- `CreateForCompiledLoop` passes the bag straight through instead of `as SharpTSObject`.
- `workerData` then structured-clones across the boundary (`StructuredClone` already supports compiled `Dictionary`/`List`/etc.) and is delivered to the worker. The worker child script always runs **interpreted** (`RunWorkerScript` news up an `Interpreter`), so it reads `workerData` via `env.Define` — the fix is purely in how the parent marshals the bag.

### Why `EmitWorkerData` is left returning null
Because worker child scripts always run interpreted, the compiled `worker_threads.workerData` path is only ever reached on the **main thread**, where Node's `workerData` is legitimately null. The PR corrects the stale comment there rather than changing behavior.

## Scope / follow-ups
- `transferList`/`resourceLimits` stay `SharpTSArray`-typed: a compiled `List<object?>` transferList yields null (so `workerData` still clones), and cross-boundary MessagePort transfer in compiled mode is a separate gap → **#406**.
- `stdin`/`stdout`/`stderr`/`resourceLimits` are parsed but inert in both modes → **#407**.

## Tests
New `WorkerThreadsTests` (both interpreter + compiled):
- `Worker_WorkerData_PrimitiveIsVisibleInWorker` — `workerData: 123` round-trips.
- `Worker_WorkerData_ObjectIsClonedAndVisibleInWorker` — object `workerData` is cloned and its fields readable.

Full `WorkerThreadsTests` suite green (62 passed).

### DIFF
--- a/Compilation/Emitters/Modules/WorkerThreadsModuleEmitter.cs
+++ b/Compilation/Emitters/Modules/WorkerThreadsModuleEmitter.cs
@@ -88,8 +88,11 @@ public sealed class WorkerThreadsModuleEmitter : IBuiltInModuleEmitter
         var ctx = emitter.Context;
         var il = ctx.IL;
 
-        // workerData is stored as a thread-local, access via runtime helper
-        // For now, return null in compiled code (workers load from file, not inline)
+        // workerData is only defined in worker context, and worker child scripts
+        // always run under the interpreter (SharpTSWorker.RunWorkerScript news up an
+        // Interpreter), which binds the cloned workerData via env.Define. This
+        // compiled path is therefore only ever reached on the MAIN thread, where
+        // Node's workerData is null — so null is the correct value here (#380).
         il.Emit(OpCodes.Ldnull);
         return true;
     }
--- a/Runtime/Types/SharpTSWorker.cs
+++ b/Runtime/Types/SharpTSWorker.cs
@@ -96,7 +96,7 @@ public class SharpTSWorker : SharpTSEventEmitter, IDisposable
     /// emitted <c>$EventLoop.Schedule</c>). Used to deliver worker events to the
     /// parent without relying on an ambient <see cref="SynchronizationContext"/>.
     /// </param>
-    public SharpTSWorker(string filename, SharpTSObject? options, Interpreter? parentInterpreter,
+    public SharpTSWorker(string filename, object? options, Interpreter? parentInterpreter,
         Action? eventLoopRef = null, Action? eventLoopUnref = null, Action<Action>? eventLoopSchedule = null)
     {
         _loopSchedule = eventLoopSchedule;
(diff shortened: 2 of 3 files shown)


nickna:
Rebased onto current `main` to reconcile with the #407 stdio cleanup that landed in parallel:
- `main` removed the dead `stdin`/`stdout`/`stderr`/`resourceLimits` fields and documented why (#407).
- This PR layers the #380 fix on top: constructor widened to `object?`, new `ReadOption` helper, and `workerData`/`transferList` marshaled through it (so they work for both the interpreter `SharpTSObject` and the compiled `Dictionary<string, object?>`). stdio/resourceLimits remain intentionally unread per #407.

The combined `WorkerThreadsTests` suite (64 tests, both modes) passes — including `main`'s `Worker_UnsupportedStdioAndResourceLimitsOptions_AreIgnored` and the two new `Worker_WorkerData_*` tests.

**Re: the earlier CI failure on this PR** — it was an unrelated flaky compiled `net.connect` test (`NetConnectPositionalPortHostCallback`), root-caused and fixed separately in #419 / PR #422 (the emitted socket wasn't Ref-ing the event loop during an in-flight connect). Not caused by this change.

Issue #482, opened 2026-06-13
FinalizationRegistry.register() type-checker signature treats first parameter as held value, not target


nickna:
Found while resolving `FinalizationRegistry<T>` type references (#456).

The type checker's `GetFinalizationRegistryMemberType` (`Runtime/BuiltIns/BuiltInTypes.cs`) models `register` as `(heldValue: T, any, any) => void` with `RequiredParams: 1` — i.e. it treats the **first** parameter as the held value `T`. But the spec and the SharpTS runtime (`SharpTSFinalizationRegistry.Register(target, heldValue, unregisterToken)`, which validates `target` is an object) take:

```ts
register(target: WeakKey, heldValue: T, unregisterToken?: WeakKey): void;
```

Consequences, both visible once the registry is strongly typed (post-#456):

- `fr.register("held")` **type-checks** but throws `Runtime Error: FinalizationRegistry target must be an object.` at runtime, because the runtime reads the first arg as the target.
- `fr.register(obj, "held")` — the **correct** call — fails type-checking against `FinalizationRegistry<string>`, because `obj` is checked against the held-value type `T = string`.

So there is no call shape that satisfies both the checker and the runtime for a typed registry.

### Fix
Correct the `register` signature in `GetFinalizationRegistryMemberType` to `register(target: object, heldValue: T, unregisterToken?: object): void` (target + optional token are `WeakKey`/object; held value is `T`). `unregister(unregisterToken: object): void` already looks right.

### Repro
```ts
let fr: FinalizationRegistry<string> = new FinalizationRegistry((v: string) => {});
fr.register("held");          // type-checks, but RUNTIME error: target must be an object
```

Issue #488, opened 2026-06-13
Generator function expression in IIFE/call position: 'yield' not recognized


nickna:
Found while implementing #456 (unrelated; surfaced when writing a generator IIFE in a scratch test).

A generator function **expression** invoked inline (an IIFE) does not establish a generator context, so `yield` in its body is rejected.

### Repro

```ts
const it = (function* () { yield 1; })();   // Type Error: 'yield' is only valid inside a generator function.
console.log(it.next().value);
```

### What works (so the `*` is parsed, but the generator context is lost in call position)

```ts
const g = function* () { yield 1; };   // OK
const it = g();
console.log(it.next().value);          // 1
```

So a generator function expression bound to a variable is fine, but the same expression parenthesized-and-called inline loses its generator-ness during type checking (the body is checked as if non-generator). Likely the function-expression-in-call-argument path doesn't set the "inside generator" flag before checking the body.

Issue #487, opened 2026-06-13
Generator<T>/AsyncGenerator<T> type references reject the lib's optional TReturn/TNext (arity fixed at 1)


nickna:
Found while implementing #456. The `Iterator`/`IterableIterator` arm added there accepts the lib's defaulted type-parameter range (`Iterator<T, TReturn = any, TNext = any>`, 1–3 args). The pre-existing `Generator` / `AsyncGenerator` arms in `ResolveGenericType` (`TypeChecker.Generics.cs`) still require **exactly 1** type argument, so they reject the lib spelling.

### Repro

```ts
function* gen(): Generator<number, void, unknown> { yield 1; }
//                ^ Type Error:  Generator requires exactly 1 type argument, got 3.
```

The lib signatures are `Generator<T = unknown, TReturn = any, TNext = unknown>` and `AsyncGenerator<T = unknown, TReturn = any, TNext = unknown>`.

### Fix shape

In `ResolveGenericType`, change the `Generator` / `AsyncGenerator` arms to accept 1–3 arguments and keep only the first (the yield type), dropping `TReturn`/`TNext` — exactly like the `Iterator` / `IterableIterator` arm added in #456. Out-of-range should be **TS2707** ("requires between 1 and 3 type arguments"), the code `tsc` uses for a defaulted range, matching the Iterator arm.

nickna:
Already resolved by 533d0cfb (which landed #548/#532/#487/#483): the `Generator`/`AsyncGenerator` arms in `ResolveGenericType` now accept 1–3 type arguments (keeping the yield type, dropping `TReturn`/`TNext`) and report TS2707 for out-of-range — exactly the requested fix shape. Verified the issue's exact repro `function* gen(): Generator<number, void, unknown> { yield 1; }` type-checks and runs in both modes. [#695](https://github.com/nickna/SharpTS/pull/695) adds explicit regression coverage (`Generator_ThreeTypeArgs_LibSpelling_Accepted`, `Generator_TooManyTypeArgs_Rejected`, `AsyncGenerator_ThreeTypeArgs_LibSpelling_Accepted`) and closes this.

Issue #486, opened 2026-06-13
Truthiness narrowing `if (x)` does not remove null/undefined from a union for member access


nickna:
Found while implementing #456 (`WeakRef<T>.deref()` returns `T | undefined`, so consuming it needs narrowing — but the gap is general and unrelated to #456).

Truthiness narrowing — `if (x) { … }` / `if (!x) return;` — does **not** remove `null`/`undefined` from a union when a member is then accessed. Explicit comparison and `typeof` narrowing work, so the gap is specific to the truthiness form.

### Repro

```ts
function f(d: string | undefined): number {
  if (d) { return d.length; }   // Type Error: Property 'length' cannot be accessed on 'undefined'. Object is possibly null or undefined.
  return 0;
}
```

Same failure for an object union (`{ n: number } | undefined`) and for a `const` bound to a call returning `T | undefined`.

### What already works (so this is narrowing-form-specific, not a missing-feature)

```ts
if (d !== undefined) { return d.length; }     // OK
if (d !== null)      { return d.n; }           // OK  (for { n } | null)
if (typeof d === "string") { return d.length; } // OK
```

`tsc` narrows `if (d)` on `string | undefined` to `string` inside the block. SharpTS should treat a truthiness guard as removing `null`/`undefined` (and falsy primitives where applicable) from the narrowed type, the same way the `!== null/undefined` and `typeof` guards already do.

PR #496, opened 2026-06-13
Fix #456: resolve IterableIterator/Iterator/WeakRef/FinalizationRegistry type references

### LINKED ISSUE #456: Built-in container type references IterableIterator/Iterator, WeakRef, FinalizationRegistry resolve to `any`
Found while fixing #347 (PR #455).

PR #455 made `Map<K,V>`, `Set<T>`, `WeakMap<K,V>`, and `WeakSet<T>` **type references** resolve to their dedicated `TypeInfo` records in `ResolveGenericType` (`TypeChecker.Generics.cs`) — previously they fell through to the `any` fallback. `Generator`/`AsyncGenerator` references already resolved.

The remaining dedicated container records are still **not** resolved from a type reference, so they degrade to `any`:

- `IterableIterator<T>` / `Iterator<T>` → `TypeInfo.Iterator` (note: this record is already *produced* by `Map`/`Set`/array `.keys()`/`.values()`/`.entries()`, but a type *annotation* `IterableIterator<T>` does not resolve to it)
- `WeakRef<T>` → `TypeInfo.WeakRef`
- `FinalizationRegistry<T>` → `TypeInfo.FinalizationRegistry`

### Impact

```ts
let it: IterableIterator<number> = [].values();
it.totallyFakeMethod(); // no type error today — `it` is `any`

type IterVal<T> = T extends IterableIterator<infer V> ? V : never;
// IterVal<...> can't bind V because the extends-side reference is `any`
```

Both strong-typing of these annotations and conditional-type `infer` extraction from them are blocked by the same missing resolution.

### Fix shape

Same as PR #455's Map/Set arms in `ResolveGenericType` (construct the record from the type arguments with a TS2314 arity check), plus add the names to `IsBuiltInGenericName` (`TypeChecker.TypeNodes.cs`). If/when added, also extend `DecomposeBuiltInContainer` (`TypeChecker.Generics.Conditional.cs`) to include them so conditional-`infer` matching works — PR #455 deliberately scoped `DecomposeBuiltInContainer` to exactly the references that resolve, so it stays non-speculative.

Open questions:
- The canonical spelling for the iterator (`Iterator<T>` vs `IterableIterator<T>` — and the lib.d.ts 3-type-param `Iterator<T, TReturn, TNext>`). Worth checking against the `tsc` lib types before committing to an arity.
- Whether `IsCompatible` / member-type lookup already covers `WeakRef`/`FinalizationRegistry` (Map/Set/WeakMap/WeakSet did; verify these before resolving them, or they'll resolve to a record with no usable members).

nickna:
Closes #456.

`IterableIterator<T>` / `Iterator<T>`, `WeakRef<T>`, and `FinalizationRegistry<T>` carry dedicated `TypeInfo` records that were only ever constructed from *expressions* (`[].values()`, `new WeakRef()`, …). A type **reference** in annotation position fell through to the `any` fallback, so the annotation was untyped and conditional-type `infer` could not extract its element. This completes the Map/Set/WeakMap/WeakSet work from #347 / PR #455, across the same four seams.

```ts
let it: IterableIterator<number> = [].values();
it.totallyFakeMethod();   // now a compile-time error (was `any` → runtime TypeError)

type ElemOf<T> = T extends IterableIterator<infer V> ? V : never;
let x: ElemOf<IterableIterator<number>> = "no";  // now errors (V binds to number)
```

### Changes
- **`ResolveGenericType`** — `Iterator`/`IterableIterator` → `TypeInfo.Iterator` (accepts the lib's `Iterator<T, TReturn = any, TNext = any>`: 1–3 args, keeps `T`, out-of-range is **TS2707**); `WeakRef`/`FinalizationRegistry` → their records (exactly 1 arg, **TS2314**).
- **`IsBuiltInGenericName`** — add the four names so the built-in shadows a like-named user alias on both resolution paths (mirrors the Map/Set precedent).
- **`DecomposeBuiltInContainer`** — add the three records so conditional `infer` binds.
- **`IsCompatible`** — these dedicated iterable records had **no compatibility arm** and fell through to the final `return false`, so even `let g: Generator<number> = gen()` already failed. Added covariant arms for `Iterator`/`Generator`/`AsyncGenerator`/`FinalizationRegistry`. A sync `Generator` satisfies an `Iterator` target, and a structurally-typed object with a callable `next` satisfies `Iterator<T>` (tsc types it structurally; exercised by `ArrayStaticTests.Array_From_CustomIterator`) — both keep positions that were `any` before from regressing.

### Rider: `FinalizationRegistry.register` signature
Corrected to `register(target: WeakKey, heldValue: T, unregisterToken?)` — `T` was mis-placed at parameter 0. Previously unobserved (the annotation was `any`); the strongly-typed annotation now makes the primary `register(target, held)` usage type-check.

### Validation
- Full suite **11479/11479**
- TypeScript conformance **31/31** (baseline unchanged)
- Test262 **0 logical failures** (its test-host crash reproduces identically on clean `main` — environmental, unrelated)
- Manual checks pass in interpreted **and** compiled modes (IL verifies)

### Follow-ups filed
- #485 — iterables modeled nominally: structural element typing, structural `for...of`, and `IteratorResult<T>`/`Iterable<T>` not modeled
- #486 — truthiness `if (x)` narrowing doesn't remove `null`/`undefined`
- #487 — `Generator<T>`/`AsyncGenerator<T>` references reject the lib's optional `TReturn`/`TNext`
- #488 — generator function expression in IIFE/call position: `yield` not recognized

### DIFF
--- a/Runtime/BuiltIns/BuiltInTypes.cs
+++ b/Runtime/BuiltIns/BuiltInTypes.cs
@@ -556,14 +556,21 @@ public static class BuiltInTypes
     }
 
     /// <summary>
-    /// FinalizationRegistry has register() and unregister() methods.
+    /// FinalizationRegistry has register() and unregister() methods. <paramref name="targetType"/> is
+    /// the registry's element type T (the held value), not the GC target.
     /// </summary>
     public static TypeInfo? GetFinalizationRegistryMemberType(string name, TypeInfo targetType)
     {
         return name switch
         {
-            "register" => new TypeInfo.Function([targetType, new TypeInfo.Any(), new TypeInfo.Any()],
-                new TypeInfo.Undefined(), RequiredParams: 1),
+            // register(target: WeakKey, heldValue: T, unregisterToken?: WeakKey): void
+            // The GC target and unregister token are arbitrary objects — SharpTS does not model the
+            // WeakKey type precisely, so `any` keeps them permissive — while heldValue carries the
+            // registry's element type T. target and heldValue are required; the token is optional.
+            // (Before #456 the FinalizationRegistry<T> annotation degraded to `any`, so this signature
+            // was never observed against an explicit T; T was previously mis-placed at parameter 0.)
+            "register" => new TypeInfo.Function([new TypeInfo.Any(), targetType, new TypeInfo.Any()],
+                new TypeInfo.Undefined(), RequiredParams: 2),
             "unregister" => new TypeInfo.Function([new TypeInfo.Any()],
                 new TypeInfo.Primitive(Parsing.TokenType.TYPE_BOOLEAN)),
             _ => null
--- a/SharpTS.Tests/TypeCheckerTests/DedicatedContainerTypeReferenceTests.cs
+++ b/SharpTS.Tests/TypeCheckerTests/DedicatedContainerTypeReferenceTests.cs
@@ -1,0 +1,334 @@
+using SharpTS.Tests.Infrastructure;
+using SharpTS.TypeSystem.Exceptions;
+using Xunit;
+
+namespace SharpTS.Tests.TypeCheckerTests;
+
+/// <summary>
+/// The single-/dual-argument dedicated built-in container records — <c>IterableIterator&lt;T&gt;</c> /
+/// <c>Iterator&lt;T&gt;</c> (=> <c>TypeInfo.Iterator</c>), <c>WeakRef&lt;T&gt;</c>,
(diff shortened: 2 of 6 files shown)


nickna:
Superseded by #484, which was merged concurrently and resolves #456 with substantially the same approach. Closing as a duplicate. The one non-duplicate piece of this branch — the FinalizationRegistry.register type-checker signature fix — is being reopened as a focused PR against current main to close #482 (which #484 left open).

Issue #506, opened 2026-06-13
Structural member collection doesn't walk a generic-instantiation superclass (CollectPublicInstanceMembers stops at InstantiatedGeneric base)


nickna:
## Summary

When a class `extends Generic<Concrete>` (its superclass is an `InstantiatedGeneric`, not a bare `TypeInfo.Class`), the structural member collectors stop walking at that base, so members inherited from a generic superclass are invisible. This affects **structural assignability** and — after #492 — **conditional `infer` matching**, both of which source members from these collectors.

```ts
class Base<T> { value: T; constructor(v: T) { this.value = v; } }
class Sub extends Base<number> {}
interface HasNum { value: number; }

// tsc: OK — Sub inherits value: number from Base<number>.
// SharpTS: error "Type 'Sub' is not assignable to type 'HasNum'".
const h: HasNum = new Sub(5);
```

```ts
// infer-match variant (post-#492):
class Base<T> { value: T; constructor(v: T) { this.value = v; } }
class Sub extends Base<number> { extra(): void {} }
type J<T> = T extends { value: infer R } ? R : "no";
// tsc: R = number, so 123 is fine. SharpTS: takes the false branch "no", so 123 errors.
let x: J<Sub> = 123;
```

Note the internal inconsistency: ordinary member **access** (`new Sub(5).value`) *does* resolve to `number` — that path (`GetMemberType` in `TypeChecker.Compatibility.Structural.cs`) walks `InstantiatedGeneric` superclasses. Only the member-set *collectors* don't.

## Root cause

`CollectPublicInstanceMembers(TypeInfo.Class)` (`TypeChecker.Compatibility.Helpers.cs`) walks with `while (current is TypeInfo.Class c) { … current = GetSuperclass(current); }`. An `InstantiatedGeneric` superclass fails the `is TypeInfo.Class` test, so the loop exits and the generic base's members are never collected. `CollectGenericClassMembers` has the mirror gap: it only folds in a `gc.Superclass is TypeInfo.Class sc` base (`extends NonGenericBase`), not a generic one (`extends Base<…>`).

## Suggested fix

[...]

Issue #505, opened 2026-06-13
declare class drops its `extends` superclass (CheckDeclareClass never resolves SuperclassExpr)


nickna:
## Summary

A `declare class` (ambient declaration — also the `@DotNetType` external-type path) **ignores its `extends` clause entirely** for the type system. `CheckDeclareClass` builds the class's `MutableClass` but never resolves `SuperclassExpr` / `SuperclassTypeArgs` nor sets `Superclass`, so inherited members are invisible to every consumer that reads a declare-class instance's member set: member-access type resolution, structural assignability, and conditional-type `infer` matching.

```ts
declare class Base { val: "correct"; }
declare class Sub extends Base {}

// tsc: s.val has type "correct"; returning it as a number is error TS2322.
// SharpTS: type-checks with NO error — `val` is not found on Sub (superclass dropped).
function f(s: Sub): number { return s.val; }
```

## Root cause

`TypeChecker.Statements.Classes.cs` `CheckDeclareClass` creates the `MutableClass` with only `IsAbstract` set:

```csharp
var mutableClass = new TypeInfo.MutableClass(classStmt.Name.Lexeme)
{
    IsAbstract = classStmt.IsAbstract
};
```

The regular `CheckClassDeclaration` path resolves the superclass first (the ~50-line block at the top of the method: `extends Box<number>` generic instantiation, `extends SomeClass` Instance/Class, `extends <Any-typed global>` placeholder, plus the `TS2507`/`TS2314`/`TS2315` error cases) and assigns `mutableClass.Superclass = superclass`. `CheckDeclareClass` performs none of this and never touches `classStmt.SuperclassExpr`.

## Impact

Affects all `declare class … extends …` declarations. Because `declare class` is how external `.NET`/ambient types are surfaced (`@DotNetType`), inherited members on those types are not visible to the checker. Concretely:

- `instance.inheritedMember` does not resolve to its declared type.
- Structural assignability against a declare-subclass omits inherited members.
- Conditional `infer` matching against a declare-subclass takes the false branch for inherited members (this is how it was found — it blocks the verbatim `declare class` repro in #492, whose own fix targets the infer-match property source and is verified with regular classes).

## Suggested fix

[...]

PR #511, opened 2026-06-13
Fix #492: conditional infer-match walks the superclass chain (and includes methods)

### LINKED ISSUE #461: infer matching: `T extends { method(): infer R }` never binds R for class instances
## Summary

In a conditional type, `T extends { method(): infer R } ? ... : ...` never binds `R` when `T` is a **class instance** whose `method` is declared as a method (not a function-typed field). The match fails and the conditional silently takes its **false** branch.

```ts
type J<T> = T extends { toJSON(): infer R } ? R : "no";
declare class MyClass { toJSON(): "correct"; }

// Expected: J<MyClass> is "correct"; assigning 42 should error "42 is not assignable to 'correct'".
// Actual:   J<MyClass> is "no" (false branch); error is "42 is not assignable to '\"no\"'".
const z: J<MyClass> = 42;
```

Both modes are affected (the match runs in the type checker, independent of node vs. string type resolution).

## Root cause

`ExtractPropertiesWithTypes` (`TypeSystem/TypeChecker.Generics.UtilityTypes.cs`) is the property source used by `CheckExtendsRecursive`'s `Record`/`Interface` branches (`TypeSystem/TypeChecker.Generics.Conditional.cs`). For a `TypeInfo.Class` it collects public **fields** (`cls.FieldTypes`) and **getters** (`cls.Getters`) but **not methods** (`cls.Methods`). So the extends-side field `toJSON` is never found on the check side, `CheckExtendsRecursive` returns `false`, and the infer placeholder `R` is left unbound.

`TypeInfo.Class` already exposes `Methods` (a `FrozenDictionary<string, TypeInfo>` of function-typed members), so the data is available.

## Suggested fix

Include public methods when extracting properties for the infer-match path. Adding methods only makes *more* extends-side fields resolvable, so it can only turn currently-failing matches into successes (it can't break an existing match). To keep the blast radius small, prefer a method-inclusive extraction used only by the two `CheckExtendsRecursive` call sites rather than changing `ExtractPropertiesWithTypes` globally (it also feeds `keyof`/mapped-type machinery, where adding methods to class key domains is a separate, broader decision).

## Notes / interaction

Found while completing #346. This bug is normally **masked**: a conditional with a method-bearing object type in its extends clause rendered with an internal arrow (`{ toJSON: () => infer R }`) was, in the string resolver, not even recognized as a conditional (see the sibling `=>` scan bug) and garbled to `any`, which is assignable to anything. Fixing that scan bug (so the conditional is correctly recognized) **exposes** this one — `microsoft/TypeScript`'s `conformance/types/conditional/inferTypes1.ts` (the `Jsonified`/`toJSON` repro) depends on it. The two should land together.

### LINKED ISSUE #492: Conditional infer-match ignores inherited class members (ExtractPropertiesWithTypes doesn't walk the superclass chain)
## Summary

Conditional-type `infer` matching against a class instance only considers the class's **own** members — it does not walk the `Superclass` chain — so an **inherited** member is invisible to a structural extends clause and the conditional silently takes its false branch.

```ts
type J<T> = T extends { val: infer R } ? R : "no";
declare class Base { val: "correct"; }
declare class Sub extends Base { extra(): void; }
type X = J<Sub>;   // expected: "correct" (val is inherited);  actual: "no" (false branch)
```

## Root cause

`ExtractPropertiesWithTypes` (`TypeSystem/TypeChecker.Generics.UtilityTypes.cs`), the property source for `CheckExtendsRecursive`'s `Record`/`Interface` branches, reads only the class's own `FieldTypes`, `Getters` (and, after #461, own `Methods`). It never merges members from `Superclass`. So `Sub` is treated as having only `extra`, and the extends-side `val`/method is not found.

This affects inherited **fields, getters, and methods** uniformly:
- fields/getters: pre-existing.
- methods: #461 added own-method extraction with the same own-only limitation (deliberately mirroring fields/getters rather than widening scope).

(Contrast: ordinary member *access* / assignability does walk the inheritance chain — see `Interpreter.Properties.cs` `EvaluateGet` and `TypeChecker` `CheckGet`. The infer-match property source is the outlier.)

## Repro (method variant)

```ts
type J<T> = T extends { toJSON(): infer R } ? R : "no";
declare class Base { toJSON(): "correct"; }
declare class Sub extends Base { extra(): void; }
type X = J<Sub>;   // expected "correct"; actual "no"
```

## Suggested fix

In the `TypeInfo.Class` case (ideally only on the method-inclusive `ExtractInferMatchProperties` path added in #461, to avoid perturbing the keyof/mapped-type callers), walk `Superclass` — chasing `Instance`/`InstantiatedGeneric` wrappers and substituting type arguments for generic bases — merging members with own-wins-over-inherited precedence. Private/protected members stay excluded (as today).

Found while completing #461.

nickna:
## What

A conditional type's structural extends clause — `T extends { val: infer R } ? R : "no"` — looks up each member on the check type. For a class instance the property source (`ExtractPropertiesWithTypes`) read only the class's **own fields and getters**: no methods, and no walk up the `Superclass` chain. So an **inherited** member, or any **method**, was invisible, the structural match failed, and the conditional silently took its **false** branch.

```ts
type J<T> = T extends { val: infer R } ? R : "no";
class Base { val: "correct" = "correct"; }
class Sub extends Base { extra(): void {} }
type X = J<Sub>;   // was "no" (false branch); now "correct" (val is inherited)
```

## Fix

Add `ExtractInferMatchProperties`, used **only** by the two `CheckExtendsRecursive` Record/Interface call sites. For a class instance it sources the full structural surface tsc matches against — public fields, getters, **and** methods, merged across the entire `Superclass` chain (derived-wins precedence; the synthetic `constructor` and private/protected members excluded), with generic-class type arguments substituted. It reuses the existing structural-compat collectors (`CollectPublicInstanceMembers` / `CollectGenericClassMembers`), so infer-matching stays **consistent** with structural assignability.

`ExtractPropertiesWithTypes` is deliberately left unchanged: it also feeds the `keyof`/mapped-type machinery, where adding methods or inherited members to a class's key domain is a separate, broader decision. Confining the change to the two infer-match call sites keeps the blast radius contained. Adding members can only turn a currently-failing structural match into a success — never the reverse — so conditionals that already resolved are unaffected.

## Scope note (#461)

#492's suggested fix is framed on "the method-inclusive `ExtractInferMatchProperties` path added in #461" — but #461 was never merged, so that path didn't exist, and #492's own repro includes a **method** variant ("affects inherited fields, getters, and methods uniformly"). The method-inclusion (#461) and the superclass walk (#492) are the **same code path**, so this single change closes both. Verified: own-method, inherited-field, inherited-getter, inherited-method, two-level inheritance, derived-shadows-inherited, and generic-receiver (`J<Box<number>>`) all resolve; private-inherited and truly-absent members correctly take the false branch.

## Tests

`SharpTS.Tests/TypeCheckerTests/InheritedInferMatchTests.cs` — 15 cases covering the above (positive + negative + exclusion + as-cast path). Tests use **regular** classes deliberately (see below).

- Full unit suite: **11565 passed, 0 failed**.
- TypeScript conformance: **31 passed, 0 failed** (no baseline shift).

## Pre-existing gaps found and filed separately

- **#505** — `declare class` drops its `extends` superclass entirely (`CheckDeclareClass` never resolves `SuperclassExpr`). This is broader than infer-matching (affects member access and assignability) and is why #492's verbatim `declare class` repro still needs that fix; this PR's tests use regular classes to exercise the member source under test without tripping it.
- **#506** — structural member collection doesn't walk a generic-instantiation superclass (`class Sub extends Base<number>`); a shared limitation of `CollectPublicInstanceMembers`/`CollectGenericClassMembers` affecting both structural assignability and (via this PR) infer-matching. Fixing it in the shared collectors repairs both together.

Closes #492. Closes #461.

### DIFF
--- a/SharpTS.Tests/TypeCheckerTests/InheritedInferMatchTests.cs
+++ b/SharpTS.Tests/TypeCheckerTests/InheritedInferMatchTests.cs
@@ -1,0 +1,228 @@
+using SharpTS.Tests.Infrastructure;
+using SharpTS.TypeSystem.Exceptions;
+using Xunit;
+
+namespace SharpTS.Tests.TypeCheckerTests;
+
+/// <summary>
+/// Conditional-type <c>infer</c> matching against a class instance must see the class's full
+/// structural surface — public fields, getters, AND methods, merged across the entire superclass
+/// chain (#461 own methods, #492 inherited members). Previously the property source
+/// (<c>ExtractPropertiesWithTypes</c>) read only a class's OWN fields and getters, so a structural
+/// extends clause like <c>{ val: infer R }</c> silently took its false branch when <c>val</c> was a
+/// method or was inherited.
+///
+/// Each test pins the resolved conditional via an annotation: if <c>J&lt;C&gt;</c> resolves to the
+/// true branch (the inferred member type) the matching-typed assignment is accepted and the
+/// mismatched one throws; if it wrongly took the false branch (<c>"no"</c>) the expectations invert.
+/// Regular (non-<c>declare</c>) classes are used deliberately — <c>declare class</c> currently drops
+/// its <c>extends</c> clause entirely (a separate, broader bug tracked outside this fix), which would
+/// mask the member-source behavior under test here.
+/// </summary>
+public class InheritedInferMatchTests
+{
+    // ---- inherited FIELD (#492) ----
+
+    [Fact]
+    public void InheritedField_ResolvesToInferredType()
+    {
+        // val is declared on Base; Sub inherits it. J<Sub> must be "correct", so "correct" assigns.
+        var source = """
+            class Base { val: "correct" = "correct"; }
+            class Sub extends Base { extra(): void {} }
+            type J<T> = T extends { val: infer R } ? R : "no";
+            let x: J<Sub> = "correct";
+            """;
+        TestHarness.RunInterpreted(source);
+    }
+
+    [Fact]
+    public void InheritedField_WrongAssignment_IsTypeError()
+    {
+        // J<Sub> is "correct" (inherited), so a number violates it.
+        var source = """
+            class Base { val: "correct" = "correct"; }
+            class Sub extends Base { extra(): void {} }
+            type J<T> = T extends { val: infer R } ? R : "no";
+            let x: J<Sub> = 42;
+            """;
+        Assert.ThrowsAny<TypeCheckException>(() => TestHarness.RunInterpreted(source));
+    }
+
+    // ---- inherited GETTER (#492) ----
+
+    [Fact]
+    public void InheritedGetter_ResolvesToInferredType()
+    {
+        var source = """
+            class Base { get val(): "correct" { return "correct"; } }
+            class Sub extends Base { extra(): void {} }
+            type J<T> = T extends { val: infer R } ? R : "no";
+            let x: J<Sub> = "correct";
+            """;
+        TestHarness.RunInterpreted(source);
+    }
+
+    // ---- inherited METHOD (#492 method variant) ----
+
+    [Fact]
+    public void InheritedMethod_ResolvesToInferredReturnType()
+    {
+        // toJSON is an inherited method; { toJSON(): infer R } must bind R to its return type.
+        var source = """
+            class Base { toJSON(): "correct" { return "correct"; } }
+            class Sub extends Base { extra(): void {} }
+            type J<T> = T extends { toJSON(): infer R } ? R : "no";
+            let x: J<Sub> = "correct";
+            """;
+        TestHarness.RunInterpreted(source);
+    }
+
+    [Fact]
+    public void InheritedMethod_WrongAssignment_IsTypeError()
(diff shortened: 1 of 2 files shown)


nickna:
### Relationship to PR #498 (also Fix #461)

PR #498 (open) addresses #461 by surfacing a class's **own** public methods (an `includeMethods` flag on `ExtractPropertiesWithTypes`, plus an `ExtractInferMatchProperties` wrapper). This PR introduces a method of the **same name** but sources the **full** structural surface — own **and inherited** fields/getters/methods, with generic-base substitution — by reusing the structural-compat collectors. It is a strict functional **superset** of #498, so the two cannot co-merge (duplicate method + identical call-site edits).

This PR's tests subsume #498's scenarios (own method incl. the `declare class` repro form, method-parameter infer, interface-method parity, private-method exclusion, no-match false branch). Recommend **closing #498 in favor of this PR**.

Issue #563, opened 2026-06-13
Compiled: function completing without an explicit return value yields `null` instead of `undefined`


nickna:
## Summary

In **compiled mode**, a plain (non-generator) function that completes **without an explicit `return <expr>`** — either a bare `return;` or falling off the end of the body — produces JS **`null`** instead of **`undefined`**. Interpreter mode is correct. Per ECMA-262, a function that returns without a value (or completes normally) has return value `undefined` (`[[Call]]` → `undefined` when the function code returns no value).

This is the **plain-function analog** of #480 (which is the interpreter generator-completion gap, now fixed). Generators are unaffected — both modes give `undefined` for a generator that completes without `return <expr>`. `return undefined;` (an explicit undefined expression) is also already correct in compiled mode; only the *implicit* completion is wrong.

## Repro

```ts
function fReturn() { return; }
function fOffEnd() {}
console.log(typeof fReturn());   // node: undefined
console.log(String(fOffEnd()));  // node: undefined
console.log(fReturn() === undefined); // node: true
```

| | bare `return;` | off-the-end | `return null;` | `return undefined;` |
|---|---|---|---|---|
| Node / interpreter | `undefined` | `undefined` | `null` | `undefined` |
| SharpTS **compiled** | **`null`** ✗ | **`null`** ✗ | `null` ✓ | `undefined` ✓ |

```
$ dotnet run -- --compile repro.ts -o repro.dll && dotnet repro.dll
typeof fReturn()  -> object   (should be undefined)
String(fOffEnd()) -> null     (should be undefined)
```

## Likely cause

When a function has no return value, the emitted IL returns a default `object` (CLR `null`) rather than the `undefined` sentinel (`$Undefined.Instance`). The fix is to materialize `undefined` (not `null`) on the no-value completion path in `EmitReturn` / the function epilogue for non-void object-returning functions. Care is needed for genuinely `void`-typed slots (see #/VoidFunctionEarlyReturn — bare `return;` in a `void` function must stay valid IL) and to keep `return null;` producing `null`.

## Discovered

While fixing #480 (interpreter generator completion value). The interpreter root-cause fix (a bare `return;` produces the `undefined` sentinel at `VisitReturn`) makes the interpreter fully correct here; this is the remaining compiled-mode half for plain functions.

nickna:
[...]

Issue #540, opened 2026-06-13
Async generator: next() on an already-completed generator replays the stale completion value (both modes)


nickna:
@
## Summary

Calling `.next()` on an **already-completed async generator** replays the stale completion/last-yielded value instead of `undefined`, in **both** the interpreter and compiled modes. Per ECMA-262 27.6.1.2 / 27.5.1.2, once a generator is in the *completed* state, every subsequent `.next()` must resolve to `{ value: undefined, done: true }`.

This is the **async-generator analogue of #499** (which fixed the *sync* compiled generator). It is **not** covered by the existing async/sync follow-ups:
- **#481** (compiled async) covers the resumed-`yield` value, the `yield*` completion value, and the off-the-end / bare-`return;` completion *value* — but **not** the re-entry-after-done replay below.
- **#480** (interpreter) is scoped to the *sync* generator types (`SharpTSGenerator` / `SharpTSArrowGenerator`), not the async-generator machinery.

## Repro

```ts
async function* ag() { yield 1; return 42; }
async function main() {
  const it = ag();
  console.log((await it.next()).value);  // 1          (ok)
  console.log((await it.next()).value);  // 42         (return value, delivered once — ok)
  console.log((await it.next()).value);  // 42  ✗      should be undefined
  console.log((await it.next()).value);  // 42  ✗      should be undefined
}
main();
```

Both **interpreter** and **compiled** print `42` for the 3rd/4th `.next()`; Node prints `undefined`. The same replay happens for the last *yielded* value after `gen.return(v)`.

A related, narrower gap (likely the same root cause on the completion path): an async generator that runs off the end reports a `null` completion value in the **interpreter**:

```ts
async function* ag2() { yield 7; }
// after consuming 7, the next result is { value: null, done: true } in interp — should be undefined
```

(The compiled off-the-end `null` is already tracked by #481; the interpreter side is not.)

## Likely cause

Mirrors #499 / #480 repro #3, one layer over in the async machinery:
[...]

Issue #489, opened 2026-06-13
Compiled generator: return null produces undefined completion value instead of null


nickna:
## Summary

In **compiled mode**, a generator's explicit `return null` produces a completion value of `undefined` instead of `null` — i.e. the compiler cannot distinguish `return null` from `return undefined` / bare `return;` for a generator's completion value. The interpreter is correct.

This is narrowly scoped: ordinary `null` literals, `null` returned from a non-generator function, `yield null`, and reading a `null`-valued object property all behave correctly in compiled mode. Only the generator *completion* value (the `value` of the final `{done: true}` result) is affected.

## Repro

```ts
function* g() { return null; }
console.log("a:" + g().next().value);
```

- Interpreter: `a:null`
- Compiled:    `a:undefined`

Counter-checks that all behave correctly in both modes (so the bug is specific to generator `return null`):
```ts
console.log("x:" + null);                       // x:null  (both)
function f() { return null; } "x:" + f();       // x:null  (both)
function* h() { yield null; } h().next().value; // null    (both)
const d: any = { value: null }; d.value;        // null    (both)
```

## Notes

- Pre-existing; discovered while fixing #443 (which is about the resumed `yield` value and the `yield*` completion value being `undefined`). This `return null` divergence is independent of that fix — `return null` takes the `r.Value != null` branch of `GeneratorMoveNextEmitter.EmitReturn`, which #443 did not touch.
- Compiled treats `return null` identically to `return undefined` / bare `return;` (all three yield an `undefined` completion value), so the null vs undefined distinction is lost somewhere on the generator return path (candidate: a return-value coercion against the inferred `TReturn`, or the done-path result construction in `GeneratorStateMachineBuilder.next()`). `yield null` storing/reading the same `CurrentField` correctly is the key clue that narrows it to the return path.
- Interpreter-side completion-value gaps are tracked separately in #480.

nickna:
Verified already fixed on `main` by c966c9a4 (which added a #489 regression test). Compiled `function* g() { return null; }` now yields a `null` completion value, distinct from `return;` / `return undefined` (both `undefined`). Coverage: `GeneratorUndefinedValueTests`. Confirmed green in a full run (0 failed / 12639 passed). Closing as completed.

PR #577, opened 2026-06-13
Fix #480: bare `return;` completes generators (and functions) with undefined

### LINKED ISSUE #480: Interpreter generator completion value is null/stale instead of undefined
## Summary

In **interpreted mode**, the *completion value* of a generator (the `value` of the final `{done: true}` iterator result) diverges from the spec in several related ways. Compiled mode was made spec-correct in #443; the interpreter still has these gaps. Note these are *not* the cases #443 covers — the interpreter already gets the resumed `yield` value and the `yield*` completion value right. The gaps below are all on the **direct `.next().value`** completion path.

Per ECMA-262, a generator that completes without an explicit `return <expr>` has completion value `undefined`, and once a generator is in the *completed* state every subsequent `.next()` returns `{value: undefined, done: true}`.

## Repros (interpreted)

**1. Off-the-end completion value is `null`, should be `undefined`:**
```ts
function* g() { yield 1; }
const it = g();
console.log("a:" + it.next().value); // a:1
console.log("b:" + it.next().value); // interp: b:null  — should be b:undefined
```
Compiled (after #443): `b:undefined` ✓

**2. Bare `return;` completion value is `null`, should be `undefined`:**
```ts
function* g() { yield 1; return; }
const it = g();
console.log("b:" + it.next().value); // interp: b:null  — should be b:undefined
```
Compiled (after #443): `b:undefined` ✓

**3. After `.return(v)` / completion, a subsequent `.next()` replays the stale value instead of `undefined` (affects BOTH modes):**
```ts
function* g() { yield 1; yield 2; }
const it = g();
it.next();                    // {value:1}
it.return(42);                // {value:42, done:true}
console.log(it.next().value); // interp: 42, compiled: 1 — should be undefined
```
```ts
function* g() { yield 1; return 7; }
const it = g();
it.next(); it.next();         // {value:7, done:true}  (return value delivered once)
console.log(it.next().value); // both: 7 — should be undefined
```

## Root cause (interpreter)

`SharpTSGenerator`/`SharpTSArrowGenerator` (`Runtime/Types/SharpTSGenerator.cs`) store `_returnValue` as CLR `null` by default and never distinguish "no return value (→ `undefined`)" from "returned `null`". `Next()` returns `new SharpTSIteratorResult(_returnValue, done: true)` on completion, and the already-completed early-return path replays `_returnValue` instead of yielding `undefined`.

A correct fix needs proper `undefined`-vs-`null` discipline (must not turn an explicit `return null` into `undefined`) plus the "once completed, always `{value: undefined}`" rule. The repeat-after-completion case (#3) is shared with compiled mode (the emitted state machine's completed-state path and `return(v)` method likewise replay the stale `Current`).

Discovered while fixing #443.

nickna:
Closes #480.

## Problem

A bare `return;` produced C# `null` (JS null) instead of the `undefined` sentinel, so an interpreter generator that completed via a value-less `return;` reported `{ value: null, done: true }` instead of `{ value: undefined, done: true }`:

```ts
function* g() { yield 1; return; }
const it = g();
it.next().value;   // 1
it.next().value;   // interp: null  →  now: undefined
```

The same root cause also made a **plain/async function**'s bare `return;` yield null rather than undefined (`function f(){ return; }; f() === undefined` was false).

Critically, the value-less `return;` is indistinguishable from `return null;` by the time it reaches the generator's completion handler (both are C# null there), so the fix must be at the source.

## Fix

`VisitReturn` / `ExecuteReturnAsyncVT` now emit the `undefined` sentinel for a value-less `return;`, while `return <expr>` is preserved verbatim — so `return null;` still yields null. The off-the-end and next-after-completion completion values were already correct in the interpreter.

## Tests

- `GeneratorUndefinedValueTests` — the two previously **compiled-only** completion-value tests (off-end, bare `return;`) are promoted to cross-mode; added a `return null;` → null guard.
- `FunctionReturnValueTests` (new) — bare `return;` / off-end → undefined, `return null;` → null, async-function bare `return;` → undefined. The implicit-completion cases are interpreter-only because compiled mode still returns null for them (filed #563).

Full suite green (11931).

## Scope / follow-ups
- #563 — compiled mode: a plain function completing without an explicit return value yields null instead of undefined (the compiled half).
- #540 — async **generators** (`SharpTSAsyncGenerator`) have the same bare-`return;`/off-end gap via their own `ExecuteStatementsAsync` handler; left to that issue to avoid overlapping its next-after-done replay work. Commented there with the exact root-cause lines.

### DIFF
--- a/Execution/Interpreter.Async.cs
+++ b/Execution/Interpreter.Async.cs
@@ -479,9 +479,10 @@ public partial class Interpreter
 
     internal async ValueTask<ExecutionResult> ExecuteReturnAsyncVT(Stmt.Return returnStmt)
     {
-        object? returnValue = null;
-        if (returnStmt.Value != null) returnValue = (await EvaluateAsync(returnStmt.Value)).ToObject();
-        return ExecutionResult.Return(returnValue);
+        // Bare `return;` completes with `undefined`, not null — see VisitReturn (#480).
+        if (returnStmt.Value == null)
+            return ExecutionResult.Return(RuntimeValue.Undefined);
+        return ExecutionResult.Return((await EvaluateAsync(returnStmt.Value)).ToObject());
     }
 
     internal async ValueTask<ExecutionResult> ExecutePrintAsyncVT(Stmt.Print printStmt)
--- a/Execution/Interpreter.cs
+++ b/Execution/Interpreter.cs
@@ -2523,9 +2523,14 @@ public partial class Interpreter : IDisposable
 
     internal ExecutionResult VisitReturn(Stmt.Return returnStmt)
     {
-        object? returnValue = null;
-        if (returnStmt.Value != null) returnValue = Evaluate(returnStmt.Value);
-        return ExecutionResult.Return(returnValue);
+        // A bare `return;` completes with `undefined` — distinct from `return null;`. Emitting the
+        // undefined sentinel here (rather than C# null, which represents JS null) is what makes a
+        // generator's completion value and a plain function's return value `undefined` instead of
+        // conflating them with null (#480). `return <expr>` preserves whatever the expression
+        // evaluates to, so an explicit `return null;` still yields null.
+        if (returnStmt.Value == null)
+            return ExecutionResult.Return(RuntimeValue.Undefined);
+        return ExecutionResult.Return(Evaluate(returnStmt.Value));
     }
 
     internal ExecutionResult VisitPrint(Stmt.Print printStmt)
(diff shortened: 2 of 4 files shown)


nickna:
Closing as superseded by #591, which is already merged to `main`.

Merging current `main` into this branch produces a tree **identical** to `main` — this PR contributes no changes. #591 ("Fix completion values 480/489/563") landed the same #480 interpreter fix as this PR, plus the compiled-mode halves (#563 compiled `return;`/off-end → `undefined`, and #489 compiled `return null` preservation) that this PR had left as `InterpretedOnly` TODOs. The bare-`return;`/off-end completion-value tests are therefore now cross-mode (`All`) on `main`, and the original CI failure here was an unrelated live-DNS smoke-test flake (since mitigated by #590).

PR #579, opened 2026-06-13
Fix labeled `continue` to an outer `for` loop (#558)

### LINKED ISSUE #558: Labeled continue to an outer for-loop re-runs the loop initializer (infinite loop), both modes
## Summary

A labeled `continue <label>` whose target is an outer `for` loop branches to a continue label placed **before** the loop's initializer, so the initializer re-runs and the loop never makes progress — an infinite loop. Reproduces in both the interpreter and the compiler, and is independent of `try`/`finally` and of generators. Found while fixing #500 (the finally itself runs correctly first — see below).

## Repro (non-generator)

```ts
const out: number[] = [];
outer: for (let i = 0; i < 2; i++) {
  for (let j = 0; j < 2; j++) {
    out.push(i * 10 + j);
    if (j === 0) continue outer;
  }
}
console.log(out.join(","));
```

Node: `0,10`. SharpTS compiled: hangs (re-runs `i = 0` forever). Interpreter: also fails to terminate normally.

## Repro (generator — same root cause, hangs both modes)

```ts
function* g(){ outer: for(let i=0;i<2;i++){ for(let j=0;j<2;j++){ yield i*10+j; if(j===0) continue outer; } } }
for (const v of g()) console.log("v" + v);
```

Node: `v0`, `v10`. SharpTS interpreter and compiled: loop forever on `v0`.

## Likely cause

`StatementEmitterBase.EmitLabeledStatement` marks the continue label *before* emitting the labeled statement, so for a `for` loop the label sits ahead of the initializer rather than at the increment. `AsyncMoveNextEmitter` overrides `EmitLabeledStatement` specifically for this ("Labeled continue for for loops (base doesn't handle correctly)"); the base, the plain `GeneratorMoveNextEmitter`, and `ILEmitter` do not, and the interpreter has an analogous gap. The fix is to point a labeled `continue` at the loop's own continue/increment target.

## Note (re #500)

With #500's fix, the enclosing `finally` *does* run on each `continue outer` (compiled now prints `v0`, `f00`, `v0`, `f00`, … matching the interpreter) — so this is purely the continue-target placement, not a finally-routing gap.

nickna:
Closes #558.

## Problem

A labeled `continue <label>` whose target is an outer `for` loop branched to a continue point placed **before** the loop's initializer, so the initializer re-ran and the loop never made progress — an infinite loop. Reproduced in **both** the interpreter and the compiler, independent of `try`/`finally` and generators.

```ts
const out: number[] = [];
outer: for (let i = 0; i < 2; i++) {
  for (let j = 0; j < 2; j++) {
    out.push(i * 10 + j);
    if (j === 0) continue outer;
  }
}
console.log(out.join(","));   // Node: 0,10 — SharpTS: hung (both modes)
```

The generator variant hung the same way (compiled looped forever on the first yielded value).

## Root cause

- `StatementEmitterBase.EmitLabeledStatement` and `ILEmitter.EmitLabeledStatement` marked the labeled continue target *before* emitting the loop. For a `for`, that sits ahead of the initializer.
- The interpreter's `ExecuteLabeledStatement` restarted the whole loop on a labeled continue (which re-runs a `for` initializer) and omitted `Stmt.For` from its loop-label set entirely.
- `AsyncMoveNextEmitter` already special-cased this; the base, the plain generator emitter, `ILEmitter`, and the interpreter did not.

## Fix

Make the loop own its label so a labeled `continue` lands on the loop's real continue point (a `for`'s increment, a `while`'s condition, an iterator's advance) — exactly like an unlabeled `continue`.

- **Compiled:** a labeled statement that directly wraps a loop parks the label on `CompilationContext.PendingLoopLabel`; the loop's own `EnterLoop` adopts it, so the loop's correctly-placed break/continue targets carry the label. `EmitForOf` emits several alternative runtime paths, so it captures the label once and hands it to every path (an early version regressed `break` on a typed-array for-of because only the first-emitted path was labeled — now fixed and covered by a test). Non-loop and chained-label statements keep the prior wrapper behavior. Covers `ILEmitter` and the sync generator emitter; the async emitter was already correct and is untouched.
- **Interpreter:** the wrapping label(s) are threaded into the loop executors, which resolve a matching labeled break/continue natively via `HandleLoopResult` instead of restarting the loop. Chained labels (`a: b: for`) all attach to the loop.

## Tests

Adds `LabeledStatementTests` cases (both interpreter and compiled) for labeled `continue`/`break` to an outer `for`/`for-of`/`for-in`/`do-while`, self-continue on a `for`, the generator repro, and chained labels. The existing suite only ever exercised labeled `while`/`for-of`/`do-while` continue — never a labeled `for` — which is why this went undetected.

## Verification

- Full unit suite: **11,890 passed**. The only 2 failures were the live-network `DnsRecordTypeTests.*ResolveNs*` smoke tests, which pass on re-run on both this branch and clean `main` (network timing, unrelated to control flow).
- Compiled IL for the labeled-loop cases passes `--verify`.
- Test262 / TypeScriptConformance: the TS-conformance suite is type-checker-only and the type checker is untouched; the Test262 subset doesn't cover labeled statements, and unlabeled loop codegen is byte-identical (the `labelName=null` path), so neither is affected.

## Known follow-up

A *chained* label on a `for` (`a: b: for ...; continue a`) is now handled by the interpreter but still hangs in compiled mode (the outer label of a chain falls back to the pre-existing wrapper placement). Filed as a separate issue; out of scope for #558 (single-label).

### DIFF
--- a/Compilation/CompilationContext.cs
+++ b/Compilation/CompilationContext.cs
@@ -170,6 +170,26 @@ public partial class CompilationContext
     // Loop control labels (with optional label name for labeled statements)
     public Stack<(Label BreakLabel, Label ContinueLabel, string? LabelName)> LoopLabels { get; } = new();
 
+    /// <summary>
+    /// Label name awaiting attachment to the next loop's own break/continue targets.
+    /// Set by <c>EmitLabeledStatement</c> when the labeled statement is a loop, and consumed
+    /// by that loop's <c>EnterLoop</c> (here or in an emitter override). This lets
+    /// <c>continue &lt;label&gt;</c> branch to the loop's real continue point — a for-loop's
+    /// increment, a while's condition — instead of a point ahead of the loop's initializer,
+    /// which would re-run it forever (#558).
+    /// </summary>
+    public string? PendingLoopLabel { get; set; }
+
+    /// <summary>
+    /// Returns the pending labeled-loop name and clears it, so it attaches to exactly one loop.
+    /// </summary>
+    public string? TakePendingLoopLabel()
+    {
+        var label = PendingLoopLabel;
+        PendingLoopLabel = null;
+        return label;
+    }
+
     // Hoisted array type caches: stack of per-loop dictionaries mapping
     // variable name → (typed local, descriptor) for arrays whose isinst
     // check has been hoisted to the loop preamble.
@@ -275,7 +295,9 @@ public partial class CompilationContext
 
     public void EnterLoop(Label breakLabel, Label continueLabel, string? labelName = null)
     {
-        LoopLabels.Push((breakLabel, continueLabel, labelName));
+        // An unlabeled EnterLoop call adopts any label parked by an enclosing labeled
+        // statement, so the loop's own continue/break targets carry the label (#558).
+        LoopLabels.Push((breakLabel, continueLabel, labelName ?? TakePendingLoopLabel()));
     }
 
     public void ExitLoop()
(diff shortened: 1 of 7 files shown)


nickna:
Known follow-up filed as #580 (compiled chained-label `continue` to the outer label of a `for` chain — interpreter handles it, out of scope for #558's single-label fix).

Issue #495, opened 2026-06-13
Test infra: live-network DNS/HTTP tests cause false-red CI (hang to 30s timeout when runner has no network)


nickna:
## Problem

A small set of tests make **real outbound network calls** (live DNS resolution and one live HTTP fetch). When the CI runner has no/slow/unreliable network — common on hosted runners — these calls don't fail fast; the in-flight request keeps the event loop alive until the **30s test timeout**, producing a false-red build. Because CI uses **fail-fast**, one such timeout on one platform **cancels the other platforms'** jobs, so the whole PR goes red.

This is **environmental flakiness, not a product bug** — but it repeatedly masks real signal and erodes trust in CI. Per the standing position on this repo (root-cause flakiness; don't just bump timeouts — see #387), these should be made **hermetic**, not given longer timeouts.

This needs to be tracked, investigated, and resolved in a dedicated session.

## Evidence (observed on PR #479, a pure parser change with zero network surface)

- CI run [27464065568](https://github.com/nickna/SharpTS/actions/runs/27464065568) went red on the **first** attempt: `HttpModuleTests.FetchReturnsPromise(mode: Interpreted)` timed out at **30s** on ubuntu (couldn't reach `example.com`); macOS + Windows were **cancelled** by fail-fast. A **re-run with no code change went fully green** — the definition of a flake.
- Locally during the same work I separately saw empty-output/timeout flakes from `DnsAsyncTests.DnsPromises_Resolve4`, `DnsRecordTypeTests.ResolveNs_InvalidDomain_Promise_Rejects`, and `DnsRecordTypeTests.LiveSmoke_ResolveNs_Promise`, each of which **passed on retry**.

## Inventory of live-network tests

**The good news: the hermetic infra already exists** and most tests use it — a local HTTP test server (`_server.BaseUrl` in `HttpModuleTests`) and a fake DNS server (`DnsFakeServerModuleTests`, `DnsWireProtocolFakeServerTests`, `DnsWireProtocolRetryTests` via `server.Address` / `fake.test`). Only these tests bypass it and hit the real network:

**Live HTTP (real internet):**
[...]

PR #596, opened 2026-06-14
Add labeled-continue async-generator regression test (closes #589)

### LINKED ISSUE #586: Compiled async generator: labeled break to an outer loop runs finally but does not exit the loop (main CI red, from #559)
## Summary

In a **compiled async generator**, a labeled `break` targeting an outer loop runs the intervening `finally` but does **not** actually exit the targeted loop — iteration continues as if the `break` were ignored. This makes `main` (`origin/main`) red: the test `AsyncGeneratorTryFinallyTests.LabeledBreakToOuterLoop_RunsInterveningFinally(mode: Compiled)` (added by PR #572 / #559) fails deterministically.

## Repro

```ts
async function* g() {
  outer: for (let i = 0; i < 3; i++) {
    for (let j = 0; j < 3; j++) {
      try { yield i * 10 + j; if (j === 1) break outer; } finally { console.log("fin" + i + j); }
    }
  }
}
async function main() { for await (const v of g()) console.log("v" + v); }
main();
```

**Expected (Node / spec):** `break outer` exits both loops after the second yield, running only the active `finally`:
```
v0
fin00
v1
fin01
```

**Compiled actual:** the `finally` runs (`fin01`) but the labeled break does not exit — iteration continues:
```
v0
fin00
v1
fin01
v2
fin02
v10
fin10
v11
...
```

## Notes

- Deterministic (not flaky) — reproduces 3/3.
- `CompiledOnly` test; the compiled async-generator state machine treats the labeled `break outer` like a no-op/continue rather than transferring control out of the labeled loop. The `finally`-dispatch half (from #559) works; the loop-exit half does not.
- `main`'s CI has been red on this since PR #572 (#559) merged (run for merge commit `d62f6675` and latest `4c86a2dc` both failed on this test).
- Sibling to #500/#558 (labeled control-flow + finally in compiled *sync* generators); this is the async-generator labeled-`break` analog.
- Encountered while merging `origin/main` into the #514 PR (#546); unrelated to #514 (interpreter `yield*`), which is green.

### LINKED ISSUE #589: Compiled async generator: labeled break to outer loop is ignored (committed test LabeledBreakToOuterLoop_RunsInterveningFinally fails on main)
## Summary

In **compiled mode**, a **labeled `break` targeting an outer loop from inside a nested loop in an async generator** is **ignored** — the loops run to completion instead of breaking out. The committed test `AsyncGeneratorTryFinallyTests.LabeledBreakToOuterLoop_RunsInterveningFinally(mode: Compiled)` currently **fails on `main`**.

This is a labeled-break analog in compiled async generators (cf. #580, which is the `continue <outerLabel>` case for a plain `for`). The test was added by #559 (PR #572); it appears a later change regressed it, or it is environment-sensitive — it reproduces deterministically here.

## Repro

```ts
async function* g() {
  outer: for (let i = 0; i < 3; i++) {
    for (let j = 0; j < 3; j++) {
      try { yield i * 10 + j; if (j === 1) break outer; } finally { console.log("fin" + i + j); }
    }
  }
}
async function main() { for await (const v of g()) console.log("v" + v); }
main();
```

- Expected: `v0`, `fin00`, `v1`, `fin01` (break outer fires at i=0,j=1, running the inner finally on the way out).
- Compiled actual: `v0`, `fin00`, `v1`, `fin01`, `v2`, `fin02`, `v10`, `fin10`, `v11`, … — `break outer` is completely ignored and both loops run to completion.

## Notes

- Pre-existing on `main` (reproduced with the working tree reverted to `origin/main`), and **not** introduced by the completion-value work in #480/#489/#563 — flagged while running the full suite during that work.
- Likely in the compiled async-generator loop/label + flag-based finally machinery (`AsyncGeneratorMoveNextEmitter*`), the same area as #558/#559. A red committed test means the suite is currently failing on `main`; worth confirming CI status.

nickna:
## Summary

[#589](https://github.com/nickna/SharpTS/issues/589) reported that the committed test `AsyncGeneratorTryFinallyTests.LabeledBreakToOuterLoop_RunsInterveningFinally` fails on `main` in **compiled mode** — a labeled `break outer` from a nested loop inside an async generator is ignored and both loops run to completion.

**The bug was already fixed by #586 (`bf3c48dc`)**, which made the compiled async generator's `EnterLoop` override adopt the pending loop label set by an `outer:`-style labeled statement (`labelName ?? Ctx.TakePendingLoopLabel()`), matching the sync generator emitter. Issue #589 was filed against a working tree that predated that merge.

This PR does **not** change production code — it closes a regression-coverage gap.

## Verification of the attribution

- Reverting #586's one line reproduces #589's **exact** reported symptom: `v0, fin00, v1, fin01, v2, fin02, v10, fin10, v11, …` (the `break outer` is a no-op).
- Restoring #586 makes the test pass **deterministically** (3/3 isolated runs). The defect is IL-emit-time label resolution, not environment-sensitive (the issue's "environment-sensitive?" hedge was the predated checkout, not flakiness).

## The coverage gap this closes

#586's commit message claimed the fix was *"covered by the existing `LabeledBreakToOuterLoop_RunsInterveningFinally` (and the labeled-continue siblings)"* — but **no async-generator labeled-`continue` test actually existed**. Coverage before this PR:

| shape | break | continue |
|---|---|---|
| sync generator (`LabeledStatementTests`) | ✔ | ✔ |
| **async generator** (`AsyncGeneratorTryFinallyTests`) | ✔ | **✗ (missing)** |

The #586 `EnterLoop` fix drives **both** break and continue (shared `FindLabeledLoopScope`), but only break had behavioral coverage, so the continue direction could regress silently.

## Changes (test-only)

- **`LabeledContinueToOuterLoop_RunsInterveningFinally`** (behavioral, `CompiledOnly`): `continue outer` must run the inner loop's `finally`, then advance the *outer* loop (skipping the rest of the inner loop). Expected `v0, fin00, v10, fin10`.
- A matching **IL-verification** `InlineData` for the labeled-continue shape.

Both are confirmed **non-vacuous**: they fail without #586 and pass with it.

## Testing

- `AsyncGeneratorTryFinallyTests`: 37/37 pass (was 35; +1 behavioral, +1 IL-verify).
- `GeneratorTryFinallyTests`, `LabeledStatementTests`, `EmitterSyncTests` (override-drift meta-test): all green (156 total).

Closes #589.

### DIFF
--- a/SharpTS.Tests/SharedTests/AsyncGeneratorTryFinallyTests.cs
+++ b/SharpTS.Tests/SharedTests/AsyncGeneratorTryFinallyTests.cs
@@ -196,6 +196,30 @@ public class AsyncGeneratorTryFinallyTests
 
     [Theory]
     [MemberData(nameof(ExecutionModes.CompiledOnly), MemberType = typeof(ExecutionModes))]
+    public void LabeledContinueToOuterLoop_RunsInterveningFinally(ExecutionMode mode)
+    {
+        // The labeled-`continue` sibling of LabeledBreakToOuterLoop (#586/#589). `continue outer` must
+        // run the inner loop's finally and then advance the *outer* loop — skipping the rest of the
+        // inner loop — rather than continuing the inner loop. The same EnterLoop pending-label adoption
+        // (#586) that resolves the labeled break drives this path; without behavioral coverage the
+        // continue direction could regress silently (the labeled-break test alone would not catch it).
+        var source = """
+            async function* g() {
+              outer: for (let i = 0; i < 2; i++) {
+                for (let j = 0; j < 3; j++) {
+                  try { yield i * 10 + j; if (j === 0) continue outer; } finally { console.log("fin" + i + j); }
+                }
+              }
+            }
+            async function main() { for await (const v of g()) console.log("v" + v); }
+            main();
+            """;
+
+        Assert.Equal("v0\nfin00\nv10\nfin10\n", TestHarness.Run(source, mode));
+    }
+
+    [Theory]
+    [MemberData(nameof(ExecutionModes.CompiledOnly), MemberType = typeof(ExecutionModes))]
     public void BreakToLoopBetweenTwoTrys_RunsOnlyInnerFinally(ExecutionMode mode)
     {
         // The break targets a loop that sits *between* two trys: only the finally inside that loop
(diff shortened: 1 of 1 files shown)


nickna:
CI note: the first run showed a macOS failure, but it was the known live-network DNS flake `DnsRecordTypeTests.LiveSmoke_ResolveNs_Promise` (real NS lookup hung → empty output; tracked by #495), **not** this change. Evidence it's unrelated:

- This PR is test-only and touches no DNS/network/runtime code.
- The new `LabeledContinueToOuterLoop_RunsInterveningFinally(mode: Compiled)` test **passed** on that same macOS run (31 ms); the DNS test was the sole failure. (Windows merely showed canceled via fail-fast.)
- Re-running the failed jobs is now green on all three platforms (macOS/ubuntu/windows).

The root-cause for the flake (compiled `dns.resolve*` should reject on a hung query rather than hang to timeout) stays in #495's scope.

PR #612, opened 2026-06-14
Fix broken main: generator compiler references removed CapturedVariables (#608×#609)

### LINKED ISSUE #470: Compiled: nested `function` declaration inside an async/generator body fails to compile ('Unhandled statement type in ILEmitter: Function')
## Summary

In **compiled mode**, a nested `function` declaration inside any **state-machine body** — an `async function`, async arrow, generator, or async generator — fails to compile with:

```
Compile Error: Unhandled statement type in ILEmitter: Function
```

Nested function declarations compile correctly inside ordinary (non-state-machine) functions, and the **interpreter handles all cases correctly**. No `.dll` is produced.

## Repro

```ts
// async function
async function outer(): Promise<number> {
  function inner(): number { return 7; }
  return inner();
}
outer().then(v => console.log(v));
```

```ts
// generator
function* gen(): Generator<number> {
  function inner(): number { return 7; }
  yield inner();
}
for (const v of gen()) console.log(v);
```

```
$ dotnet run -- --compile repro.ts -o repro.dll
Compile Error: Unhandled statement type in ILEmitter: Function

$ dotnet run -- repro.ts      # interpreter
7
```

A nested function declaration in a **non-async** function compiles and runs fine:

```ts
function outer(): number {
  function inner(): number { return 7; }   // OK
  return inner();
}
console.log(outer());   // 7
```

## Cause

The synchronous `ILEmitter.EmitStatement` has a `case Stmt.Function:` arm (`Compilation/ILEmitter.cs:225`) — this is why CLAUDE.md's "inner function declarations are supported in the IL compiler" holds for ordinary functions. The shared `StatementEmitterBase.EmitStatement`, used by **every** state-machine emitter (`AsyncMoveNextEmitter`, `AsyncArrowMoveNextEmitter`, `GeneratorMoveNextEmitter`, `AsyncGeneratorMoveNextEmitter`), has **no** `Stmt.Function` arm, so a nested function declaration falls through to the `throw` at `Compilation/StatementEmitterBase.cs:294`.

## Scope

Compiled mode only; all four state-machine contexts (async function, async arrow, generator, async generator). Interpreter is correct. Found while adding regression tests for #457 (a parity test used a nested helper inside an `async function`).

### LINKED ISSUE #541: Compiled generators capture closure variables by value (snapshot at creation), not by reference
## Summary

Compiled generators snapshot the **values** of captured outer (closure) variables at generator-creation time instead of reading them by reference when the body runs. This diverges from the interpreter and from JS closure semantics. Found while working on #521 — it is why the compiled #521 repro surfaces `undefined is not a function` (the self-assigned `it` is captured as `undefined`).

## Repro

```ts
let x: any = 1;
function* g() { yield x; }
const it = g();   // generator created while x === 1
x = 42;           // x mutated after creation, before the body runs
console.log(it.next().value);
```

- **Node / SharpTS interpreter:** `42` (closures capture the variable, not a copy of its value).
- **SharpTS (compiled):** `1`.

Self-reference variant (the shape in #521's repro):

```ts
let it: any;
function* g() { yield (it === undefined ? "undefined" : "generator"); }
it = g();
console.log(it.next().value);   // interp: "generator"   compiled: "undefined"
```

## Notes

- Non-generator functions and arrows capture correctly by reference (`let x=1; const f=()=>x; x=42; f()` → `42` in both modes), so this is specific to the generator state-machine capture path: `GeneratorStateMachineBuilder`'s hoisted/captured fields are populated by value when the stub creates the state machine.
- Fix likely needs generators to share the enclosing scope's variable cell (the way arrow display classes do) rather than copying the value into a state-machine field at creation.
- Consequence for #521: the `let it; it = g(); it.next()` repro can't reach the new re-entrancy guard because `it` is `undefined` inside the body. Genuine re-entrancy (receiver resolved live, e.g. through an object property) does hit the guard and now throws `TypeError: Generator is already running` correctly.

nickna:
## Problem

`main` (2a5046de) **does not compile**:

```
Compilation/ILCompiler.Generators.cs(70,23): error CS1061:
  'GeneratorStateAnalyzer.GeneratorFunctionAnalysis' does not contain a
  definition for 'CapturedVariables'
```

This blocks every branch and PR, not just the one where it surfaced.

## Root cause — a semantic merge conflict git didn't flag

Two PRs touched the same area in **different files**, so the merge had no textual conflict:

- **#608** (`ef5329c3`, *Fix #470/#501: lift non-capturing nested generators*) added `ExcludeTopLevelFunctionsFromCaptures()`, which reads `analysis.CapturedVariables`. Written against a record that still had that member.
- **#609** (`3ba5789b`, *Fix #541: generators capture closure variables by reference*) **removed `CapturedVariables`** from `GeneratorFunctionAnalysis`, switching generators to read captured outer variables live-by-reference (`CapturedTopLevelVars` / `TopLevelStaticVars`), mirroring async generators.

#608 merged first (`019b320b`), then #609 (`2a5046de`) removed the member #608's new code depends on → broken `main`.

## Fix

Under #541's model the state machine no longer materializes per-variable fields from `analysis.CapturedVariables`, so the whole exclusion chain is **dead code** — its only consumer fed `DefineStateMachine`, which no longer reads captured variables:

- `_topLevelFunctionNamesByModule` (field)
- `CollectTopLevelFunctionNames()` (populator, called from 2 sites in `ILCompiler.cs`)
- `ExcludeTopLevelFunctionsFromCaptures()` (consumer)

Removed the chain and pass `analysis` straight through. The *"object is not a function"* case #608 guarded against can no longer occur (no analysis-driven captured fields exist) and is now handled by #541's unified live-reference path. The #470/#501 fix itself — the `NestedFunctionLifter` — is untouched.

Net: **3 insertions, 67 deletions**, 2 files.

## Verification

- Build clean (0 errors).
- **1703 targeted tests pass**: `NestedFunctionLifting` (#470/#501), all `Generator`/`AsyncGenerator`/`Yield`, `Closure`/`Capture`, `ILVerification`.

### DIFF
--- a/Compilation/ILCompiler.Generators.cs
+++ b/Compilation/ILCompiler.Generators.cs
@@ -12,68 +12,6 @@ namespace SharpTS.Compilation;
 public partial class ILCompiler
 {
     /// <summary>
-    /// Top-level (module-scope) function names, keyed by module path ("" for single-file).
-    /// A generator references a top-level function by treating its name as a captured outer variable
-    /// (it is not declared inside the generator). Generators wrongly materialise such captures as
-    /// state-machine fields that the kickoff never populates for functions — leaving them null, so a
-    /// value-use (<c>const h = helper; h()</c> / <c>yield helper</c>) reads null and throws "object
-    /// is not a function". Async/async-generator bodies don't create these fields and instead resolve
-    /// the name through the normal function-value path. Excluding such names from the generator's
-    /// captured fields lets its MoveNext resolver fall through to <c>TryEmitGlobalVariable</c>'s
-    /// function path too.
-    ///
-    /// <para>Keyed PER MODULE, not unioned: a name that is a top-level function in one module may be
-    /// a genuinely-captured binding (e.g. a module-level <c>const</c>, whose captured field IS
-    /// populated) in another. Excluding it there would strip a needed field. Only a name that is a
-    /// top-level function in the GENERATOR'S OWN module resolves correctly through the function-value
-    /// fallback, so only those are excluded.</para>
-    /// </summary>
-    private readonly Dictionary<string, HashSet<string>> _topLevelFunctionNamesByModule = new();
-
-    /// <summary>
-    /// Records the top-level function names in <paramref name="statements"/> under
-    /// <paramref name="moduleKey"/> (<c>""</c> for single-file). Called once per module after
-    /// nested-function lifting, so relocated functions (now top-level) are included.
-    /// </summary>
-    private void CollectTopLevelFunctionNames(IEnumerable<Stmt> statements, string moduleKey)
-    {
-        if (!_topLevelFunctionNamesByModule.TryGetValue(moduleKey, out var set))
-        {
-            set = new HashSet<string>();
-            _topLevelFunctionNamesByModule[moduleKey] = set;
-        }
-        foreach (var stmt in statements)
-        {
-            switch (stmt)
-            {
-                case Stmt.Function f when f.Body != null:
-                    set.Add(f.Name.Lexeme);
-                    break;
-                case Stmt.Export { Declaration: Stmt.Function ef } when ef.Body != null:
-                    set.Add(ef.Name.Lexeme);
-                    break;
-            }
-        }
-    }
-
-    /// <summary>
-    /// Returns <paramref name="analysis"/> with the current module's top-level function names removed
-    /// from its captured variables, so the generator state machine does not define (never-populated)
-    /// fields for them. See <see cref="_topLevelFunctionNamesByModule"/>.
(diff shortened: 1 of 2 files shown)


nickna:
Superseded by #611 (commit 165112b8, merged via d6271c6e), which landed the identical fix first — same two files, same `3 insertions / 67 deletions` dropping the dead `_topLevelFunctionNamesByModule` / `CollectTopLevelFunctionNames` / `ExcludeTopLevelFunctionsFromCaptures` chain. `main` (d6271c6e) builds clean (0 errors), so this PR is now a no-op duplicate. Closing.

Issue #625, opened 2026-06-14
Compiled: async arrow that writes to a captured variable emits invalid IL (StackUnexpected)


nickna:
## Summary

In **compiled mode**, an **async arrow** that *writes* to a variable captured from an enclosing scope emits invalid IL — `StackUnexpected` in `<>c__AsyncArrow_N.MoveNext`. **Reading** a captured variable is fine; a **non-async** arrow that writes the same variable is fine. The emitted IL still runs on the JIT (which is more lenient than ILVerify) and produces the correct result in simple cases, but it fails `--verify` and can drop the side effect in more complex state machines (see Notes).

## Repro

```ts
async function main() {
  let n = 0;
  const w = async () => { n = 5; };   // write to captured `n`
  await w();
  console.log(n);
}
main();
```

`dotnet run -- --compile repro.ts --verify` →
```
[IL Error] <>c__AsyncArrow_0.MoveNext: StackUnexpected - Unexpected type on the stack.
```
Runtime: prints `5` (correct — JIT accepts the IL). Interpreter: `5`.

Compound writes (`order += "a"`) reproduce identically.

## Contrast (both clean under --verify)

```ts
// async arrow READING a captured var — fine
async function main() { let s = "x"; const r = async () => { console.log(s); }; await r(); }

// NON-async arrow writing a captured var — fine
function main() { let n = 0; const w = () => { n = 5; }; w(); console.log(n); }
```

So the defect is specific to the **async-arrow MoveNext × captured-variable write** combination.

## Notes

- Distinct from #615 (async arrow *nesting* another async arrow → `SelfBoxedField not set`). This one has no nesting and the symptom is `StackUnexpected`.
- Pre-existing on `main`; discovered while fixing #614 (the captured-write async arrow was used as an evaluation-order probe). It is independent of #614 — it reproduces with no string method, optional chain, or call-argument await involved.
- Likely the captured-variable store path in the async-arrow `MoveNext` emitter leaves a value (the display-class/`this` reference, or the stored value) on the IL stack in a shape the verifier rejects at the store. In one longer async function the captured write was silently dropped at runtime (the JIT-accepted-but-malformed store did not persist), so this is not purely cosmetic.

nickna:
[...]

Issue #441, opened 2026-06-13
Compiled: awaiting a function-call result inside an async arrow emits invalid IL (StackUnexpected)


nickna:
## Summary

In **compiled mode**, `await <functionCall>` inside an **async arrow** emits invalid IL (`StackUnexpected` in `<>c__AsyncArrow_0.MoveNext`). The same `await <call>` shape in a regular `async function` is fine, and awaiting an inline `new Promise(...)` (rather than a call) in an async arrow is also fine — so this is specific to the **async-arrow MoveNext × await-of-call** combination. The emitted IL still runs on the JIT (which is more lenient than ILVerify), so programs produce correct output but fail `--verify`.

Discovered while fixing #414 (async-arrow spill across await). It is independent of spills — it reproduces with no value spilled across the await.

## Repro

```ts
function defer(v: any, ms: number): Promise<any> { return new Promise(r => setTimeout(() => r(v), ms)); }
const f = async () => { const x = await defer(1, 5); console.log(x); };
f();
```

`dotnet run -- --compile repro.ts --verify` →
```
[IL Error] <>c__AsyncArrow_0.MoveNext: StackUnexpected - Unexpected type on the stack.
```
Runtime: prints `1` (correct — JIT accepts the IL).

## Contrast (all clean under --verify)

```ts
// regular async function — await of a call is fine
async function m() { const x = await defer(1, 5); console.log(x); }   // 0 errors

// async arrow — await of an inline new Promise is fine
const g = async () => { const x = await new Promise<number>(r => setTimeout(() => r(1), 5)); console.log(x); };  // 0 errors
```

## Notes

- Distinct from #436 / #439 (await *in a call argument*) and #430 (async-generator `for await…of` consumption).
- Likely the async-arrow `EmitAwait`'s evaluation of the awaited call leaves the IL stack in a shape the verifier rejects on the suspend/resume merge. The regular-async-function path (`AsyncMoveNextEmitter.EmitAwait`) handles the same shape correctly, so comparing the two emitters is the place to start.

nickna:
Fixed in `main` by PR #616 (merged); the issue simply never auto-closed.

Verified the repro on current `main`:

```ts
function defer(v: any, ms: number): Promise<any> { return new Promise(r => setTimeout(() => r(v), ms)); }
const f = async () => { const x = await defer(1, 5); console.log(x); };
f();
```

- Interpreter: `1`
[...]

Issue #439, opened 2026-06-13
Compiled: await in a method-call argument (after an earlier arg) emits invalid IL / crashes — static, instance, and optional-chain dispatch


nickna:
## Summary

In **compiled mode**, calling a **method** with an `await` in an argument that follows an earlier (non-awaited) argument produces invalid IL and crashes at runtime. This affects the static, instance, and optional-chain method-dispatch paths. The interpreter is correct.

This is the **method-call** sibling of #413 (rest/spread free-function args) and #436 (fixed-arity free-function args): the dispatch paths in `ExpressionEmitterBase.CallHelpers.cs` emit each argument directly onto the IL evaluation stack (or into an inline-built args array) via `EmitExpression`, so an `await` inside a later argument suspends while the receiver and earlier args are still stacked — invalid IL. `obj.method(x, await p)` is a very common pattern, so this is impactful.

## Repros (each crashes / fails `--verify`)

**Instance method** (`TryEmitDirectMethodCall`):
```ts
function defer<T>(v: T, ms = 5): Promise<T> { return new Promise(r => setTimeout(() => r(v), ms)); }
class C { m(a: string, b: string): string { return a + b; } }
async function main() { const c = new C(); console.log(c.m("A", await defer("B"))); }
main();
```
- Interpreter: `AB`
- Compiled: `Unhandled exception. System.NullReferenceException ... at <main>d__0.MoveNext()`
- `--verify`: `<main>d__0.MoveNext: StackUnexpected` (x2)

**Static method** (`TryEmitGetCalleeViaBaseClass` static-method branch):
```ts
class C { static m(a: string, b: string): string { return a + b; } }
async function main() { console.log(C.m("A", await defer("B"))); }
```
- `--verify`: `StackUnexpected` + `PathStackDepth`; crashes at runtime.

**Optional-chain method** (`TryEmitOptionalChainMethodCall`):
```ts
const o: any = { m(a: string, b: string): string { return a + b; } };
async function main() { console.log(o?.m("A", await defer("B"))); }
```
- `--verify`: `PathStackDepth`; crashes at runtime.

(In all three, `defer` resolves from a later event-loop turn, so the state machine genuinely suspends. A single awaited arg with nothing before it, e.g. `c.m(await p)`, is unaffected.)

## Likely fix

[...]

Issue #400, opened 2026-06-13
Compiled: awaiting a deferred thenable inside an expression drops the pending stack value


nickna:
## Summary

In **compiled mode**, `await`-ing a thenable that settles its callback *later* (a deferred/macrotask resolution) from inside a larger expression loses any value already pushed on the IL evaluation stack before the `await`. The classic symptom is a string-concatenation prefix vanishing. The interpreter is correct.

This looks like an await stack-spill gap (cf. #251 "SpillBoxed pattern") that is only exercised when the awaitable resumes from a *deferred* thenable settlement rather than an already-completed Task/synchronous thenable.

## Repro

```ts
class Thenable {
  v: any = undefined;
  settled = false;
  cb: any = undefined;
  constructor(executor: (res: any, rej: any) => void) {
    executor(
      (x: any) => { this.v = x; this.settled = true; if (this.cb) this.cb(x); },
      (_e: any) => {});
  }
  then(onF: any) { if (this.settled) onF(this.v); else this.cb = onF; }
}
class P extends Promise<number> {
  static get [Symbol.species]() { return Thenable as any; }
}
async function main() {
  const r: any = P.resolve(5).then((x: number) => x + 1);
  console.log("concat: " + (await r));   // deferred thenable awaited inside a concat
}
main();
```

### Interpreter (correct)
```
concat: 6
```

### Compiled (wrong)
```
6
```

The `"concat: "` literal — pushed on the stack before the `await` — is lost when the state machine resumes from the deferred thenable callback.

## Scope / characterization

- Works fine (both modes) when the awaitable is a plain `Promise` or a **synchronous** thenable (one whose `then` calls `onF` immediately): `console.log("x: " + (await Promise.resolve(7)))` → `x: 7`.
- Breaks only when the thenable **defers** the callback and is settled from a later turn of the event loop (here, the general NewPromiseCapability `Settle` continuation posted to the event-loop `SynchronizationContext`).
- Affects any deferred thenable, not just Promise species — but it is easy to hit via `then`/`catch`/`finally` on a Promise subclass whose `@@species` is a general (non-Promise) constructor (#349). The existing #349 tests don't catch it because they use bare `console.log(await r)` with nothing pending on the stack.

## Discovered via

Surfaced while implementing #390 (function-valued / general Promise `@@species`). The #390 result-construction itself is correct; this is purely the await-resumption spill.

PR #626, opened 2026-06-14
Fix #614: emit await args once before the string-method dispatch split

### LINKED ISSUE #614: Compiled: await in an optional-chain dispatchable-string-method argument crashes at compile time
## Summary

In **compiled mode**, an `await` in an argument of an **optional-chain call to a dispatchable string method** (`substring`, `substr`, `charAt`, …) crashes at **compile time** with `Error: The given key '1' was not present in the dictionary.`. The interpreter is correct.

This is a sibling of #439 (await in a method-call argument), but in a path #439's fix did not cover: the runtime string-method fast path (`EmitRuntimeStringMethod`, reached from `TryEmitOptionalChainMethodCall`'s string branch and from `EmitDynamicMethodCallPreservingThis`) builds its argument `object[]` **inline on the IL stack** and is not await-aware, so the await-state bookkeeping for the suspended argument is never set up.

## Repro

```ts
function defer<T>(v: T, ms = 5): Promise<T> { return new Promise(r => setTimeout(() => r(v), ms)); }
const s: any = "hello";
async function main() { console.log(s?.substring(await defer(1), 4)); }
main();
```

- Interpreter: `ell` (correct)
- `--compile repro.ts`: `Error: The given key '1' was not present in the dictionary.`

The **non**-string-dispatchable optional-chain path is fine (fixed in #439): e.g. `s?.slice(await defer(1), 4)` compiles, verifies, and runs. Only the `IsRuntimeDispatchableStringMethod` names (`substring`/`substr`/`charAt`/`charCodeAt`/`toUpperCase`/`toLowerCase`/`trim`/`trimStart`/`trimEnd`/`startsWith`/`endsWith`/`repeat`/`padStart`/`padEnd`/`at`) hit the inline-args fast path.

## Likely fix

Make `EmitRuntimeStringMethod` (and the `EmitGetPropertyAndInvoke` it falls back to) spill arguments to await-safe locals before assembling the `object[]`, the same approach used for the general dispatch paths in #439 / #413. `EmitDynamicMethodCallPreservingThis` (the non-optional dynamic string dispatch, e.g. for CJS-imported receivers) shares `EmitRuntimeStringMethod` and has the same latent gap.

## Notes

Pre-existing on `main` (confirmed by stashing the #439 fix — the same crash reproduces). Discovered while implementing #439.

nickna:
## Summary

Fixes **#614**: in compiled mode, an `await` (or `yield`) in an argument of a call to a **runtime-dispatchable string method** (`substring`, `substr`, `charAt`, …) on an `any`-typed receiver crashed the **compiler** with `Error: The given key '1' was not present in the dictionary.` The interpreter was always correct.

### Root cause

These calls emit two mutually-exclusive paths selected at runtime by `isinst string`: the direct string fast path (`EmitRuntimeStringMethod`) and the generic `GetProperty`+`InvokeMethodValue` path. **Both contained the call arguments at emit time**, so a single AST `await` was emitted **twice**. `EmitAwait` assigns state numbers with a sequential counter (`_currentAwaitState++`) and looks up `_stateLabels[stateNumber]`; the analyzer registers exactly one label per AST await. The second emission advanced the counter to a state that was never registered → `KeyNotFoundException` at compile time.

This is the path #439’s fix did not cover (the issue is the *dispatchable-string* fast path specifically); `s?.slice(await …)` already worked because `slice` isn’t dispatchable and only the generic path emits its args.

### Fix

Spill each argument to an await-safe local **exactly once, before the string-vs-generic branch**, and have both branches read those locals. `SpillStoreObject`-registered spills survive the `MoveNext` re-entry (#400), and emitting the args once keeps the analyzer/emitter await counters in sync.

- **`TryEmitOptionalChainMethodCall`** — for dispatchable-string methods, pre-spill before the branch. Non-string methods keep the #439 spill-after-lookup order (**byte-identical IL**), so evaluation order is unchanged on that path.
- **`EmitDynamicMethodCallPreservingThis`** (non-optional dynamic dispatch, e.g. CJS-imported receivers) — same dual-emission gap, fixed the same way; the receiver now spills via `SpillStoreObject` so it survives an await in a later arg. This was the latent sibling gap noted in #614.
- **`EmitGetPropertyAndInvoke`** — made await-safe on its own: resolves the method to a local first (lookup precedes arg evaluation, per spec), then spills args off a clear stack.
- New **`EmitBoxedArgsArray`** helper centralises the boxed `object[]` build (inline or from pre-spilled locals), replacing five copies of the loop.

### Tests

- `AwaitDeferredSpillTests` — behavior in **both** modes: optional-chain `substring` (the repro), two awaits, the `charAt` default-case fallback, non-optional dynamic dispatch, a non-string receiver with its own `substring`, and nullish-receiver short-circuit.
- `ILVerificationTests` — verify-only guard covering the optional-chain (direct + default) and dynamic-dispatch shapes.
- Manually verified parity (interpreter == compiled, `--verify` clean) for async functions, async arrows, and **generators** (`yield` in the arg).

**Full suite: 12376 passed, 0 failed.**

## Note on #439 / #441

This branch was opened to also complete **#439** and **#441**, but both are **already fixed in `main`** by the merged PR #616 (they simply never auto-closed). Verified each repro: interpreter, `--compile --verify`, and the compiled run all pass. Closing those issues separately with evidence; no code here is needed for them.

## Follow-up filed

- **#625** — pre-existing, unrelated: an async **arrow** that *writes* to a captured variable fails IL verify (`StackUnexpected`). Reads and non-async arrows are fine. Discovered while probing evaluation order; independent of this path.

Closes #614.

### DIFF
--- a/Compilation/ExpressionEmitterBase.CallHelpers.cs
+++ b/Compilation/ExpressionEmitterBase.CallHelpers.cs
@@ -339,8 +339,9 @@ public abstract partial class ExpressionEmitterBase
     {
         EmitExpression(obj);
         EnsureBoxed();
-        var objLocal = IL.DeclareLocal(Types.Object);
-        IL.Emit(OpCodes.Stloc, objLocal);
+        // SpillStoreObject (not a plain local) so the receiver survives a MoveNext re-entry when a
+        // later argument suspends (#400/#614); in the synchronous emitter it is just a plain local.
+        var objLocal = _helpers.SpillStoreObject();
 
         // Fast path: if receiver is a string at runtime, dispatch to the string-strategy path
         // directly. GetProperty(str, methodName) returns null for everything except "length",
@@ -349,17 +350,25 @@ public abstract partial class ExpressionEmitterBase
         // receiver is a string.
         if (IsRuntimeDispatchableStringMethod(methodName))
         {
+            // The string fast path and the generic GetProperty path both contain the arguments and
+            // are mutually exclusive at runtime. When an argument can suspend, spill the arguments
+            // once here so each `await`/`yield` is emitted exactly once — emitting them in both
+            // branches would emit the suspension twice and desync the await-state counter (#614).
+            LocalBuilder[]? argLocals = AnyContainsSuspension(arguments)
+                ? arguments.Select(SpillBoxed).ToArray()
+                : null;
+
             var notStringLabel = IL.DefineLabel();
             var endLabel = IL.DefineLabel();
             IL.Emit(OpCodes.Ldloc, objLocal);
             IL.Emit(OpCodes.Isinst, Types.String);
             IL.Emit(OpCodes.Brfalse, notStringLabel);
 
-            EmitRuntimeStringMethod(objLocal, methodName, arguments);
+            EmitRuntimeStringMethod(objLocal, methodName, arguments, argLocals);
             IL.Emit(OpCodes.Br, endLabel);
 
             IL.MarkLabel(notStringLabel);
-            EmitGetPropertyAndInvoke(objLocal, methodName, arguments);
+            EmitGetPropertyAndInvoke(objLocal, methodName, arguments, argLocals);
             IL.MarkLabel(endLabel);
             SetStackUnknown();
             return;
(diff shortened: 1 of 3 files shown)


nickna:
Additional follow-up filed: **#627** — a narrow, documented parity nuance of this fix. For an optional-chain call to a dispatchable-string-method *name* on a non-string object that lacks that method, a side-effecting `await` arg is evaluated in compiled mode but skipped by the interpreter (result value `undefined` is identical in both). Emitting the `await` exactly once requires spilling before the `isinst string` split, so the args are evaluated before the generic path’s fn-nullish short-circuit. Both modes are already non-spec-conformant for this case; aligning them cleanly is tracked in #627. The non-dispatchable-name path (#439) is unaffected and stays consistent.

Issue #642, opened 2026-06-14
Interpreter: NaN === NaN returns true inside an async function (strict equality NaN gap)


nickna:
## Summary

In **interpreted** mode, strict equality of `NaN` evaluated **inside an `async` function** returns the wrong result: `NaN === NaN` is `true` (and `NaN !== NaN` is `false`), when per ECMA-262 7.2.16 it must be `false`/`true`. The same expression is correct at top level, inside a `function*` generator, and in **compiled** mode.

Discovered while adding regression tests for #600 (the compiled side was fixed there — compiled async now correctly returns `false`).

## Repro

```ts
async function main() {
  console.log(NaN === NaN);     // interp: true   (BUG — should be false)
  const x = NaN;
  console.log(x === x);         // interp: true   (BUG — should be false)
  console.log(NaN !== NaN);     // interp: false  (BUG — should be true)
}
main();
```

```
# interpreter
true
true
false

# compiled (correct)
false
false
true
```

## Scope

| Context | `NaN === NaN` (interp) |
|---|---|
| top-level / sync function | `false` ✅ |
| `function*` generator body | `false` ✅ |
| **`async function` body** | **`true` ❌** |
| compiled (any context) | `false` ✅ |

So the defect is specific to the **interpreter's async execution path**. The sync interpreter path special-cases NaN in strict equality; the async path appears to route strict equality through a comparison that honors the .NET `Double.Equals` convention (`NaN.Equals(NaN) == true`) instead of the JS `===` rule. Worth auditing where the async interpreter evaluates `Expr.Binary` strict equality vs. the sync path.

## Notes

- Pre-existing; independent of #600 (which only touched the compiled backend).
- Compiled async strict-equality NaN handling is correct as of the #600 fix (routes through the NaN-aware `EmitObjectEqualsBoxed_NoBox`).

nickna:
Resolved on `main` by 1052b6e8 ("Fix four interpreter-mode conformance bugs (#642, #603, #633, #446)").

The interpreter's async binary fast path (`EvaluateBinaryAsync` in `Execution/Interpreter.Async.cs`) compared numbers with `Double.Equals` (NaN-blind) instead of IEEE-754 `==`; it now mirrors the synchronous `EvaluateBinary` fast path. That commit also added the cross-mode regression suite `AsyncNaNStrictEqualityTests`.

Verified green on current `main` (interpreted mode):

```
async function main() {
  console.log(NaN === NaN);   // false
  const x = NaN;
[...]

Issue #641, opened 2026-06-14
Compiled: capturing async arrow nested directly inside a top-level async arrow is unsupported


nickna:
## Summary

In **compiled** mode, an async arrow nested directly inside a **top-level async arrow** that **captures** a variable from the enclosing arrow's scope is not supported. It now produces a clear compile error:

```
Compile Error: Compiled mode does not yet support a capturing async arrow nested directly inside
a top-level async arrow. Move the inner arrow to a named binding or a regular async function as a
workaround.
```

The interpreter handles it correctly. The **non-capturing** case (including parameterized inner arrows) works in both modes after #615.

## Repro

```ts
const f = async () => {
  const base = 100;
  const x = await (async () => base + 5)();   // inner arrow captures `base`
  console.log(x);                              // interp: 105; compiled: compile error
};
f();
```

Also affects capture-through-await chains:
```ts
const f = async () => {
  let n = 1;
  const g = async () => { n = await (async () => n + 10)(); };  // captures `n`
  await g();
  console.log(n);  // interp: 11
};
```

## Cause / scope

A top-level async arrow's nested async arrows are registered as **standalone** state machines (`DefineTopLevelAsyncArrows`). #615 made `EmitNestedAsyncArrow` emit a standalone nested async arrow as a plain `TSFunction(null, stub)` — correct when the inner arrow has no captures (its stub only takes its own params). When the inner arrow captures enclosing locals, those values must be threaded into the standalone stub's capture args (the `StandaloneCaptureFields`), analogous to `EmitCapturingArrowInAsyncArrow` for non-async arrows. That wiring isn't in place, so the emit deliberately errors rather than passing the wrong target (which previously crashed with `InvalidCastException` / produced wrong results).

The corresponding case for arrows nested inside async **functions** already works (they're set up non-standalone with the function's boxed state machine as the outer reference).

## Workaround

Move the inner arrow to a named binding or a regular `async function`:
```ts
const f = async () => {
  const base = 100;
  const inner = async () => base + 5;   // named — compiles fine
  const x = await inner();
  console.log(x);
};
```

## Discovered

While fixing #615 (non-capturing nested async arrow failing to compile).

nickna:
[...]

Issue #640, opened 2026-06-14
Compiled: cross-module/value-call omitted optional args pad as CLR null, not the `undefined` sentinel


nickna:
## Summary

In **compiled** mode, calling a function **as a value** with fewer arguments than parameters pads the missing trailing optional parameters with CLR `null` instead of the `undefined` sentinel. This is observable through `typeof`, `=== undefined`, and `=== null`. The interpreter is correct. **Same-module direct calls are correct** (they pad with the sentinel) — only the value/dynamic-dispatch path (cross-module imports, callbacks, `$TSFunction.Invoke`) is affected.

This was masked until #600: compiled state-machine `=== undefined` previously treated null as undefined, so a null-padded arg compared `=== undefined` as true. With #600 fixed (strict equality is now correct), the null padding is exposed.

## Repro

```ts
// helper.ts
export function h(x?: any): string { return typeof x; }

// main.ts
import { h } from './helper';
console.log(h());            // interp: "undefined"   compiled: "object"  (BUG)
console.log(h() === undefined); // interp: true        compiled: false      (BUG)
```

Same-module is fine:
```ts
function g(x?: any): string { return typeof x; }
console.log(g());  // both: "undefined"
```

## Cause

`$TSFunction.AdjustArgs` (`Compilation/RuntimeEmitter.TSFunction.cs`, standard padding path ~line 1506-1529) pads missing args via `new object[paramCount]` whose unfilled slots are CLR `null`. The inline comment documents this as **deliberate**: emitted built-ins broadly use null-checks for optional-arg absence (stream write/end callbacks, encodings, …), so sentinel-padding would make them treat an absent slot as a present argument.

The tension: built-in methods invoked via `$TSFunction` want `null` padding, but **user TypeScript functions** want the `undefined` sentinel (JS semantics). Both share `AdjustArgs`, so a correct fix must distinguish the two (e.g. a per-$TSFunction `padWithUndefined` flag set when wrapping a user method vs a runtime built-in).

## Impact

- `typeof` / `=== undefined` / `=== null` give wrong answers for omitted optional params across module boundaries.
[...]

Issue #615, opened 2026-06-14
Compiled: async arrow nesting another async arrow fails to compile (SelfBoxedField not set)


nickna:
## Summary

In **compiled mode**, an **async arrow** whose body contains a **nested async-arrow expression** (e.g. an immediately-invoked async arrow) fails to compile with:

```
Compile Error: Async arrow with nested arrows does not have SelfBoxedField set.
```

The interpreter is correct. The same nested async-arrow inside a regular `async function` compiles fine, and an async arrow awaiting a *named* async arrow compiles fine — so this is specific to **an async arrow that directly nests another async-arrow expression**.

The `<>__selfBoxed` mechanism (#414) lets a nested async arrow capture the one shared state-machine instance of its enclosing async function/arrow. It is provisioned when the enclosing scope is an async function, but not when the enclosing scope is itself an async arrow, so the nested-arrow emit hits the unset-field guard.

## Repro

```ts
const f = async () => { const x = await (async () => 9)(); console.log(x); };
f();
```

- Interpreter: `9` (correct)
- `--compile repro.ts`: `Compile Error: Async arrow with nested arrows does not have SelfBoxedField set.`

## Contrast (both compile cleanly)

```ts
// outer is a regular async function — nested async-arrow IIFE is fine
async function f() { const x = await (async () => 9)(); console.log(x); }

// outer arrow awaits a *named* async arrow (no directly-nested arrow expression) — fine
const inner = async () => 9;
const f2 = async () => { const x = await inner(); console.log(x); };
```

## Likely fix

Provision the self-boxed field (and the `EmitSelfBoxedStartAndReturnTask` kickoff, `StateMachineEmitHelpers`) for an async **arrow** that itself contains nested async arrows, mirroring the async-function path in `ILCompiler.Async.cs` / `AsyncArrowStateMachineBuilder`.

## Notes

Pre-existing on `main` (unrelated to the literal-emission area). Discovered while investigating #441.

PR #644, opened 2026-06-14
Fix #568, #573, #600, #615: compiled slot types + async/generator undefined & strict equality

### LINKED ISSUE #568: Compiled: `T | undefined` parameter reassigned to `undefined` crashes (String slot / invalid IL)
Found while adding a compiled-parity test for #556, but **pre-existing and unrelated to #556** — it reproduces with no narrowing involved at all. Interpreter handles every form correctly; only the IL backend is affected.

A parameter whose declared type is `T | undefined` is compiled into a non-nullable slot for `T`. Reassigning `undefined` to it (the `$Undefined` sentinel object, which is *not* a CLR null) then fails.

### Repro 1 — `string | undefined` → `InvalidCastException`

```ts
function f(x: string | undefined): number {
  x = undefined;
  return 1;
}
console.log(f("hi"));
```

Compiled output:
```
Unhandled exception. System.InvalidCastException: Unable to cast object of type '$Undefined' to type 'System.String'.
   at $Program.f(String)
```

The compiled signature is `f(String)` — the `string | undefined` parameter resolved to a non-nullable `String` slot, so storing/reading the `$Undefined` sentinel throws.

### Repro 2 — `number | undefined` → invalid IL / CLR crash

```ts
function f(x: number | undefined): number {
  x = undefined;
  return 1;
}
console.log(f(3));
```

Compiled output: `Fatal error. Internal CLR error. (0x80131506)` (the value-type analogue produces unverifiable IL rather than a clean cast failure).

### What works (scoping)

- `string | null` parameter reassigned to `null` → **works** (CLR null is a valid `String` reference; only the `$Undefined` sentinel is the problem).
- A **local** `let x: string | undefined = s` reassigned to `undefined` → **works** (locals get an `object` slot). Only **parameters** are mis-typed.
- A `string | undefined` parameter that is only **read** (e.g. `if (x !== undefined) return x.length;`) → **works** (object slot).

### Likely root cause

Parameter slot type resolution for `T | undefined` reference/value unions, in the same family as the param-reassign coercion work in #402/#372 (`EmitConvertForParamSlot` before `Starg`). A `T | undefined` (or any union whose runtime values include the `$Undefined` sentinel) parameter should use an `object` slot — as locals already do — or the store path must box/convert the sentinel like locals do.

### Notes
- `--verify` would likely flag Repro 2.
- Discovered via `WhileCondition_NarrowsUseBeforeLaterReassignment_Compiled` in `ControlFlowNarrowingTests`; that test now uses a local to avoid this bug while still exercising #556 in compiled mode.

### LINKED ISSUE #629: Compiled state machines emit the undefined literal as CLR null -> throw undefined / any undefined operand stringifies as 'null'
## Summary

In **compiled** code, when an `undefined` literal appears as a value operand inside a state machine (async function, generator, async generator, async arrow), it is emitted as CLR `null` (`Ldnull`) instead of the `$Undefined` sentinel. The value still compares `=== undefined` (the runtime treats CLR null as undefined for `===`), but it **stringifies as `"null"`**, so any code that renders it is wrong. Non-state-machine code (the main `ILEmitter` path) is correct.

Found while fixing #619: a `throw undefined` whose `catch` now runs (after the #619 gate fix) prints `caught e=null` in compiled async/generator code instead of `caught e=undefined`.

## Root cause

`undefined` as an expression routes through `ExpressionEmitterBase.EmitUndefinedConstant()` → `StateMachineEmitHelpers.EmitUndefinedConstant()`:

```csharp
public void EmitUndefinedConstant()
{
    if (_runtime != null)
        _il.Emit(OpCodes.Ldsfld, _runtime.UndefinedInstance);   // correct: $Undefined sentinel
    else
        _il.Emit(OpCodes.Ldnull);                               // fallback: CLR null
}
```

The state-machine emitters construct `StateMachineEmitHelpers` **without** the `EmittedRuntime` (it is not known at the emitter's construction time — only `_ctx.Runtime` becomes available later, during `EmitMoveNext*`), so `_runtime` is null and the `Ldnull` fallback is always taken. The same emitters do emit `_ctx.Runtime.UndefinedInstance` directly elsewhere (completion values, etc.), so this only bites `undefined` used as a general operand.

## Repro

```ts
async function f(){
  try { throw undefined; } catch(e){ console.log("u=" + e + " isUndef=" + (e === undefined)); }
  try { throw null; }      catch(e){ console.log("n=" + e + " isNull="  + (e === null)); }
}
f();
```

| | Output |
|---|---|
| Node / interpreter | `u=undefined isUndef=true`, `n=null isNull=true` |
| SharpTS compiled | `u=null isUndef=true` ❌, `n=null isNull=true` |

(The same `u=null` appears for a `throw undefined` literal inside a compiled generator/async-generator `try/catch`.)

## Likely fix

Make the state-machine `EmitUndefinedConstant` resolve the `$Undefined` sentinel via the now-available `_ctx.Runtime.UndefinedInstance` rather than falling back to `Ldnull` — e.g. wire the runtime into `StateMachineEmitHelpers` once `_ctx` is set at the start of `EmitMoveNext*`, or override `EmitUndefinedConstant` in the state-machine emitters. Blast radius is every `undefined` literal in every state machine, so it warrants its own change with full suite + Test262 validation (kept out of the #619 fix for that reason).

nickna:
Closes #568, #573, #600, #615. Also fixes #629 (same root cause as #600).

All four are compiled-mode IL-backend bugs; the interpreter was correct in each case. Each fix ships with both-mode (or compile-and-verify) regression tests. Full xUnit suite: **12387 passing, 0 failing** (+11 new). Test262 default subset: no baseline diff. TypeScriptConformance is unaffected (type checker untouched).

## #600 — `await` of a null-resolved promise compared `=== undefined` as true (also #629)
Three coordinated root causes inside state-machine bodies (async function, async arrow, generator, async generator):
1. The base `ExpressionEmitterBase.EmitLiteral` had no `SharpTSUndefined` arm, so the `undefined` literal fell through to `ldnull` — CLR null — in every state machine (ILEmitter already had the arm). This is **#629** (undefined operands also stringified as `"null"`).
2. The state-machine helper was built before the `CompilationContext`/runtime was available, leaving `EmitUndefinedConstant` on its null-fallback and numeric coercion/comparison on inferior inline IL. Added `StateMachineEmitHelpers.SetRuntime`, called once `ctx` is bound.
3. `TryEmitBinaryOperator` used the null≡undefined-collapsing loose `Equals` for both `==` and `===`. Strict `===`/`!==` now route through the NaN-aware `EmitObjectEqualsBoxed` helpers, matching the non-async path.

The timers exposure (`await setTimeout(10) === undefined`) traced to omitted optional args being padded as CLR null (intentional, for built-in null-checks — **#640**). Rather than change that, the `timers/promises` stdlib re-export now declares explicit `= undefined` defaults so a missing `value` resolves `undefined`.

## #573 — Date/RegExp/Map/Set param, return & field slots
These map (strictly) to `DateTime`/`Regex`/`Dictionary`/`HashSet`, but their runtime values are dynamic `$TSDate`/`$RegExp`/`$Map`/`$Set` carried as `object` — so a typed slot failed ILVerify (`StackUnexpected`) and threw `InvalidCastException`. The return resolver already widened collections to object; added `TypeMapper.IsDynamicRuntimeType` (collections + DateTime + Regex) and applied it to **parameter** slots, **return** slots (now also Date/RegExp), and class **field**/parameter-property backing fields. Array params now use an object slot too; the remaining `arr[i]` body-index ILVerify gap is pre-existing (**#445**).

## #568 — `T | undefined` parameter slot
A `string | undefined` param compiled to a `String` slot (can't hold the `$Undefined` sentinel → `InvalidCastException`); `number | undefined` to `double?` (unverifiable IL). New `CoerceParamSlotType` falls back to `object` for unions admitting `undefined`, `Nullable<T>`, and `symbol` — mirroring locals and the return resolver.

## #615 — async arrow nesting another async arrow
A standalone async arrow nesting an async-arrow expression (e.g. an IIFE) failed to compile (`SelfBoxedField not set`). `DefineStateMachineStandalone` now provisions `<>__selfBoxed` when the arrow has nested async children (detected from the collected-arrow parent map). A nested arrow that *captures* enclosing variables would need its captures threaded through the shared state machine, which the standalone path doesn't wire — that case now fails loudly with a clear message instead of silently miscompiling (it never compiled before; tracked in **#641**).

## Issues filed
- **#642** — interpreter `NaN === NaN` returns `true` inside an `async` function (separate pre-existing bug found while testing; compiled mode is correct after #600).

### DIFF
--- a/Compilation/AsyncArrowMoveNextEmitter.ArrowFunctions.cs
+++ b/Compilation/AsyncArrowMoveNextEmitter.ArrowFunctions.cs
@@ -102,11 +102,38 @@ public partial class AsyncArrowMoveNextEmitter
                 "Nested async arrow function not registered with state machine builder.");
         }
 
-        // For nested arrows, we need to pass the current arrow's boxed state machine
-        // as the "outer" reference for the nested arrow.
-        // The nested arrow's stub expects (outer state machine boxed, params...)
+        // A nested async-arrow expression inside a top-level (standalone) async arrow — e.g. an
+        // immediately-invoked `(async () => 9)()` — is registered as its own standalone builder
+        // whose stub takes only its own parameters (and captures): there is no enclosing state
+        // machine to thread. Emit it as a plain TSFunction over that stub with a NULL target,
+        // exactly like a non-capturing arrow value. Threading the enclosing arrow's boxed state
+        // machine here instead (the nested-in-async-function mechanism) would prepend it as the
+        // stub's arg0, clobbering the first real parameter. (#615)
+        if (nestedBuilder.IsStandalone)
+        {
+            // A capturing standalone nested arrow would need its capture fields populated from the
+            // enclosing frame, which this path does not wire up — fail loudly rather than read
+            // uninitialized captures (it did not compile at all before #615; tracked in #641).
+            if (nestedBuilder.StandaloneCaptureFields.Count > 0)
+            {
+                throw new CompileException(
+                    "A nested async arrow that captures variables inside a top-level async arrow is " +
+                    "not yet supported in compiled mode; hoist it to a named async arrow or async " +
+                    "function, or run in interpreted mode.");
+            }
 
-        // Load the current arrow's self-boxed reference
+            _il.Emit(OpCodes.Ldnull);
+            _il.Emit(OpCodes.Ldtoken, nestedBuilder.StubMethod);
+            _il.Emit(OpCodes.Call, Types.MethodBaseGetMethodFromHandle);
+            _il.Emit(OpCodes.Castclass, typeof(MethodInfo));
+            _il.Emit(OpCodes.Newobj, _ctx!.Runtime!.TSFunctionCtor);
+            SetStackUnknown();
+            return;
+        }
+
+        // Non-standalone nested arrow (nested inside an async function): the nested arrow's stub
+        // expects (outer state machine boxed, params...), so thread the enclosing arrow's boxed
+        // self-reference as the "outer" target.
         if (_builder.SelfBoxedField != null)
         {
             _il.Emit(OpCodes.Ldarg_0);
@@ -114,7 +141,6 @@ public partial class AsyncArrowMoveNextEmitter
         }
         else
         {
-            // Fallback: this shouldn't happen if hasNestedAsyncArrows was set correctly
             throw new CompileException(
                 "Async arrow with nested arrows does not have SelfBoxedField set.");
         }
--- a/Compilation/AsyncArrowMoveNextEmitter.cs
+++ b/Compilation/AsyncArrowMoveNextEmitter.cs
@@ -131,6 +131,9 @@ public partial class AsyncArrowMoveNextEmitter : StatementEmitterBase, IEmitterC
     {
         // Note: _il is initialized in constructor via GetILGenerator()
         _ctx = ctx;
+        // Wire the runtime into the helper now that the context is bound, so MoveNext uses the
+        // $Undefined sentinel and JS-spec coercion/comparison helpers rather than null/Convert (#600).
+        if (ctx.Runtime != null) _helpers.SetRuntime(ctx.Runtime);
 
         // Create variable resolver for hoisted fields, locals, and captured variables
         _resolver = new AsyncArrowVariableResolver(_il, _builder, _locals);
--- a/Compilation/AsyncArrowStateMachineBuilder.cs
+++ b/Compilation/AsyncArrowStateMachineBuilder.cs
@@ -239,7 +239,9 @@ public class AsyncArrowStateMachineBuilder
             [_types.IAsyncStateMachine]
         );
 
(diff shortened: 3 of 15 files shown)


nickna:
Note: #643 is a parallel implementation of the same four issues from an earlier session. The two reach the same conclusion on #615 (a standalone arrow's nested async-arrow is emitted as an independent `TSFunction(null, stub)` — provisioning `SelfBoxedField` instead clobbers the nested arrow's first parameter). Main difference: for the #600 timers fallout (omitted optional args padding as CLR null, #640), this PR keeps the `timers/promises` *_DefaultValue tests passing in **both** modes by giving the stdlib re-export explicit `= undefined` defaults, rather than restricting them to interpreted mode. Reviewers should pick one PR to land.

Issue #547, opened 2026-06-13
Compiled generator: for-in loop containing a yield terminates after the first key (iteration state not hoisted)


nickna:
## Summary

In **compiled mode**, a `for-in` loop inside a generator that `yield`s in its body iterates only **once** — it emits the first key, then terminates instead of continuing to the remaining keys. The interpreter is correct. `for-of` in the same position works, because the generator emitter hoists the for-of enumerator to a state-machine field; `for-in` has no equivalent, so its key-enumeration position lives in an IL local that the `MoveNext` re-entry wipes.

## Repro

```ts
// (1) for-in OUTSIDE a generator — works in compiled mode
function plainForIn() {
  const obj = { a: 1, b: 2, c: 3 };
  const out: string[] = [];
  for (const k in obj) { out.push(k); }
  return out.join(",");
}
console.log("plain", plainForIn());            // a,b,c   (both modes)

// (2) for-OF with yield in a generator — works (contrast)
function* ofGen() {
  for (const x of [10, 20, 30]) { yield x; }
}
console.log("ofGen", [...ofGen()].join(","));  // 10,20,30 (both modes)

// (3) for-IN with yield in a generator — BUG
function* inGen() {
  const obj = { a: 1, b: 2, c: 3 };
  for (const k in obj) { yield k; }
}
console.log("inGen", [...inGen()].join(","));  // interp: a,b,c   compiled: a   ← BUG
```

- Interpreter: `inGen` → `a,b,c`
- Compiled:    `inGen` → `a` (stops after the first key)

A `yield`ing for-in also breaks the two-way accumulator pattern over object keys (the generator runs off the end after the first yield, so subsequent `next(v).value` are `undefined`).

## Likely cause

`GeneratorMoveNextEmitter` overrides `EmitForOf` (`Compilation/GeneratorMoveNextEmitter.Statements.cs`) to store the enumerator in a hoisted state-machine field (`GetEnumeratorField`) so it survives the `MoveNext` re-entry across a yield. There is **no** corresponding `EmitForIn` override — `EmitForIn` is inherited from `StatementEmitterBase` (see the note at the bottom of that file), so the for-in key list and current index are kept in IL locals that are wiped when `MoveNext` re-enters on the loop back-edge. The fix is to give for-in the same hoisted-iteration-state treatment for-of already has when the loop body contains a yield (the analyzer would also need a for-in analog of `ForOfLoopsWithYield`).

## Notes

[...]

Issue #697, opened 2026-06-16
Compiled: for await…of INSIDE an async generator still blocks on a genuinely-async source (AsyncGeneratorMoveNextEmitter.EmitForAwaitOf)


nickna:
## Summary

#631 made `for await…of` in an **async function** (and direct `await it.next()`) suspend instead of blocking. The sibling path — `for await…of` **inside an async generator** (`AsyncGeneratorMoveNextEmitter.EmitForAwaitOf`) — was not converted and still drives the inner iterator with a synchronous `GetResult()`. So a stream-transform async generator over a genuinely-async source blocks/crashes in compiled mode.

## Repro

```ts
function later(n: number): Promise<number> { return new Promise(res => setTimeout(() => res(n), 5)); }
async function* src() { yield await later(1); yield await later(2); }
async function* transform() { for await (const x of src()) yield x * 10; }  // for-await inside an async generator
async function main() { for await (const v of transform()) console.log("v" + v); }
main();
```

| | Result |
|---|---|
| Node | `v10`, `v20` |
| SharpTS interpreted | `Runtime Error: Cannot iterate over non-iterable value` (separate interp gap) |
| SharpTS compiled | unhandled exception in `<main>d__0.MoveNext` |

(A `for await` over an *already-settled* source inside an async generator works.)

## Likely cause

`Compilation/AsyncGeneratorMoveNextEmitter.Statements.cs` `EmitForAwaitOf` calls `next()` then `TaskAwaiter<object>.GetResult()` synchronously ("get result synchronously for now — full async continuation would require state machine suspension"), blocking the event-loop thread the inner iterator's continuation needs — the same deadlock #631 fixed in `AsyncMoveNextEmitter.EmitForAwaitOf`.

## Fix sketch

Mirror the #631 async-function fix in the async-generator emitter: have `AsyncGeneratorStateAnalyzer` reserve suspension states for the for-await's implicit `next()`/`return()` awaits, and rewrite `EmitForAwaitOf` to suspend via the async-generator's await mechanism (`EmitAwaitSuspensionReturn` / `AsyncGeneratorAwaitContinue`) instead of blocking. The interpreter's separate "cannot iterate" failure for `for await` over an async generator is a distinct gap (its eager-drain async-generator model).

## Notes
- Same root cause as #631 (synchronous drive), different emitter; left as follow-up to keep the #631 rework scoped to the async-function consumer + direct `next()` (its filed repro).

Issue #691, opened 2026-06-16
Compiled: synchronous throw in a for-await body that also awaits (or breaks/continues), inside a try, escapes the catch


nickna:
## Summary

In **compiled mode**, a `try { for await (…) { … } }` where the loop **body** both contains an `await` (or a top-level `break`/`continue`) **and** raises a *synchronous* `throw` lets that throw escape the enclosing `try` instead of reaching the `catch`. Bodies that are await-free and jump-free are handled correctly.

## Repro

```ts
async function* g() { yield 1; yield 2; }
async function main() {
  try {
    for await (const x of g()) {
      await Promise.resolve(0);          // body contains an await
      if (x === 2) throw "body error";   // synchronous throw
    }
  } catch (e) { console.log("caught " + e); }
}
main();
```

- Interpreter: `caught body error`
- Compiled: unhandled exception escapes `main`'s state machine

## Background / cause

The `for await…of` suspension fix (#631) made the loop drive the iterator through the async state machine, so the loop can no longer sit inside a real IL `try` (its resume labels would be branched into) — `try { for await … }` takes the flag-based path. To still route a body throw to the catch, `AsyncMoveNextEmitter.EmitForAwaitBody` wraps the body in a real IL try **only when the body is await-free and jump-free** (an await would re-introduce the branch-into-try; a `break`/`continue` would emit a `Br` out of the try, which the async emitter doesn't currently turn into a `Leave`). A body with an await or top-level break/continue keeps the plain path, so a *synchronous* throw in it is not captured.

Proper fix: make the async emitter exception-block-aware for break/continue (emit `Leave`), and segment the for-await body around its awaits (as `EmitTryBodyWithAwaits` does for a top-level try body), so synchronous spans are protected while awaits suspend.

## Notes
- Narrow: a rejected `await`/`next()` inside such a body *is* still caught (suspension-based path); only a *synchronous* `throw` in an await-or-jump-containing body escapes. Before #631 the same program hung (for-await over any async iterator blocked), so this is a new shape of a pre-existing gap, not a behavior that previously worked end-to-end.

Issue #690, opened 2026-06-16
Interpreter: concurrent next() on an async generator throws 'Only instances and objects have properties'


nickna:
## Summary

In the **interpreter**, issuing two `next()` calls on an async generator before the first settles (e.g. `Promise.all([it.next(), it.next()])`) throws `Runtime Error: Only instances and objects have properties`. Compiled mode services them FIFO (after #542).

## Repro

```ts
function later(n: number): Promise<number> { return new Promise(res => setTimeout(() => res(n), 5)); }
async function* g() { yield await later(1); yield await later(2); }
async function main() {
  const it: any = g();
  const [a, b] = await Promise.all([it.next(), it.next()]);
  console.log(a.value, a.done, b.value, b.done);
}
main();
```

- Interpreter: `Runtime Error: Only instances and objects have properties.`
- Compiled: `1 false 2 false` (correct, #542)

## Likely cause

The interpreter's `SharpTSAsyncGenerator` uses an eager full-body-drain model (`Runtime/Types/SharpTSAsyncGenerator.cs`) that doesn't implement the ECMA-262 §27.6.3 request queue; a second concurrent `next()` returns a value the surrounding `Promise.all` / destructuring can't treat as an iterator result. Interpreter-only.

## Notes
- Discovered while fixing #542 (the compiled request-queue fix). The interpreter's async-generator model diverges from spec more broadly (eager drain); this is one observable symptom.

nickna:
Verified fixed on `main` (7023541a + rewrite 766e5f70). `Promise.all([it.next(), it.next()])` on an async generator returns `1 false 2 false` instead of throwing. Checked 2026-06-17 (interpreter).

PR #702, opened 2026-06-16
Compiled generator/async-generator fixes: #547, #631, #542, #583

nickna:
Fixes four compiled-mode generator / async-generator issues. All are compiled-only — the interpreter uses an independent model and is unaffected. Each issue is a separate commit.

## #547 — for-in with `yield` stops after the first key
The generator MoveNext emitter hoisted the for-**of** enumerator across yields but kept the for-**in** key list + index in IL locals, which a `MoveNext` re-entry wipes. Mirror the for-of treatment: `GeneratorStateAnalyzer` records for-in loops whose body yields, `HoistingManager` allocates a key-list field + index field per such loop, and `GeneratorMoveNextEmitter.EmitForIn` reads/writes both from those fields.

## #631 + #542 — truly-async, serialized async-generator `next()`
The compiled async generator drove its body synchronously (`MoveNextAsync().AsTask().GetResult()`), so a not-yet-settled await deadlocked the event-loop thread (#631), and concurrent/re-entrant `next()` recursed into `MoveNextAsync` (only kept from overflowing by a reject-on-busy guard, never queued — #542).

- `next()` no longer blocks: it drives one step and returns the `Task<object>` built by an emitted `AsyncGeneratorBuildResult` helper.
- Requests serialize on a per-generator `_pendingTail` task chain (ECMA-262 §27.6.3 AsyncGeneratorQueue modeled as a Task chain) — overlapping `next()` calls run FIFO. A *synchronously* re-entrant `next()` still rejects with a TypeError to avoid a stack overflow (the self-reentrant case would deadlock under a real queue anyway).
- `for await…of` now suspends on the iterator protocol via the async state machine instead of blocking (`AsyncStateAnalyzer` reserves a suspension state for each implicit `next()`/`return()` await; a `try` enclosing a `for await` takes the flag-based path so its resume labels aren't branched into).
- `AsyncGeneratorAwaitContinue` no longer short-circuits on the awaited task's fault, so a pending rejection reaches the body's own resume point and its `try/catch` (also unblocks the pending sub-case of #617).

## #583 (part 1) — lift capturing nested generator/async functions
Two layers:
1. Calling a **capturing arrow inside a generator** threw "Non-static method requires a target" (the sync generator emitter bound a null `$TSFunction` target and never built the display instance). `GeneratorMoveNextEmitter` now has the capturing-arrow handling the async-generator emitter already has.
2. `NestedFunctionLifter` now lambda-lifts **inside-function** captures (was module-block only): the captured function-scope bindings become leading parameters of the relocated top-level declaration, forwarded by an in-place arrow. Self-recursive declarations decline cleanly (a compiled arrow snapshots captures by value, and the forwarding arrow's own `let` is in TDZ at creation). #583 §2 (relocated identity shared across enclosing-call instances) remains.

## Tests
- New shared (interp+compiled) tests for each fix in `GeneratorTests`, `AsyncGeneratorTests`, `GeneratorClosureCaptureTests`; the emitter-sync guardrail (`EmitterSyncTests`) updated for the two new `GeneratorMoveNextEmitter` overrides.
- Full `SharpTS.Tests` suite green (12702 passing, 0 failing). Standalone-DLL tests green (the new emitted helpers use only BCL types).

## Follow-up gaps filed while here
- #688 compiled `yield*` to a genuinely-async iterator still blocks (same synchronous-drive root, async yield* loop).
- #689 / #690 interpreter async-generator gaps (var scope in `for await` body; concurrent `next()`).
- #691 compiled: a *synchronous* throw in a `for await` body that also awaits/breaks, inside a `try`, escapes the catch (narrow tradeoff of the #631 for-await fix).
- #697 compiled `for await…of` **inside** an async generator still blocks on a genuinely-async source (`AsyncGeneratorMoveNextEmitter.EmitForAwaitOf` — same root cause as #631, separate emitter; left as follow-up to keep this rework scoped to #631's filed repro).

### DIFF
--- a/Compilation/AsyncGeneratorStateMachineBuilder.cs
+++ b/Compilation/AsyncGeneratorStateMachineBuilder.cs
@@ -47,10 +47,24 @@ public class AsyncGeneratorStateMachineBuilder
     // Flag set by return() to trigger finally blocks during MoveNextAsync resume
     public FieldBuilder ReturnRequestedField { get; private set; } = null!;
 
-    // Re-entrancy flag: true only while next()/return() synchronously drives MoveNextAsync. A guest
-    // next()/return()/throw() observing it means the generator body is advancing itself; without the
-    // guard the synchronous drive recurses into MoveNextAsync until the stack overflows (#542).
+    // Re-entrancy flag: true only while the generator body is synchronously advancing (the window of a
+    // MoveNextAsync call before it suspends or completes). A guest next()/return()/throw() observing it
+    // means the body is advancing itself; without the guard that synchronous re-entry recurses into
+    // MoveNextAsync until the stack overflows (#542). Set/cleared by <DriveOnce> and return()'s drive.
     public FieldBuilder ExecutingField { get; private set; } = null!;
+
+    // Tail of the request chain (ECMA-262 §27.6.3 AsyncGeneratorQueue, modeled as a Task chain): the
+    // Task<object> of the most recently enqueued next(). A new next() awaits this before driving, so
+    // overlapping next() calls run in FIFO order instead of re-entering MoveNextAsync concurrently and
+    // corrupting state. Lets next() be truly asynchronous — it never blocks the event-loop thread on a
+    // pending await (#631) — and serializes concurrent next() requests (#542). Null until the first next().
+    public FieldBuilder PendingTailField { get; private set; } = null!;
+
+    // <DriveOnce>(): drives one MoveNextAsync step (under the executing guard) and returns its
+    // { value, done } Task<object>; <DriveContinuation>(antecedent, state): static trampoline that
+    // calls <DriveOnce> on a queued next() once the prior request settles.
+    private MethodBuilder _driveOnceMethod = null!;
+    private MethodBuilder _driveContinuationMethod = null!;
 
     // Constructor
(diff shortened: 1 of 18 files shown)


nickna:
Rebased onto current `main` (now MERGEABLE). Two integration notes after the rebase:

- **#631 for-await**: `main`'s for-await shared-lowering refactor (#430/#645, `2b7fc821`) landed after my branch point, moving the lowering to `StatementEmitterBase.EmitForAwaitOf`. My #631 suspension fix is kept as a `protected override` of that shared base in `AsyncMoveNextEmitter` — async functions suspend; the shared base still blocks for async arrows and async generators (the latter tracked by #697).
- **#583 layer 1** (capturing arrow inside a generator) is already fixed on `main` by #674 (the base `EmitArrowFunction` now instantiates the display class). My pre-rebase `GeneratorMoveNextEmitter` arrow override was dropped as redundant; this PR's #583 contribution is just the `NestedFunctionLifter` lambda-lift extension for inside-function captures, which builds on #674.

Full `SharpTS.Tests` suite re-run on the rebased state is in progress; the targeted async/generator/closure + emitter-sync suites (264 tests) and all manual repros are green.

Issue #792, opened 2026-06-16
Compiled: a defaulted generator/async-generator parameter captured by a nested arrow ignores its default on omitted args


nickna:
## Summary

In **compiled** mode, when a **generator** or **async-generator** parameter has a **default value** *and* is **captured-and-mutated by a nested arrow/closure**, an **omitted** argument does not get its default: the `undefined` sentinel leaks into the closure's storage instead. The interpreter is correct. Supplying the argument explicitly works in both modes — only the default-firing (omitted) path is broken, and only when the parameter is captured.

Affects generators and async-generators, instance methods and free functions, and both value-type and reference-type defaults.

## Repro

```ts
// value-type default, captured by arrow — free-function generator
function* g(acc: number = 5): Generator<number> { [1, 2, 3].forEach(n => acc += n); yield acc; }
console.log([...g()][0]);             // interp: 11    compiled: NaN

// reference-type default, captured — instance generator method
class C { *gen(s: string = "x"): Generator<string> { [1, 2, 3].forEach(n => s += n); yield s; } }
console.log([...new C().gen()][0]);   // interp: x123  compiled: 6

// value-type default, captured — async generator
async function* ag(acc: number = 5) { [1, 2, 3].forEach(n => acc += n); yield acc; }
(async () => { for await (const v of ag()) console.log(v); })();  // interp: 11  compiled: NaN
```

| case | interp | compiled |
|------|--------|----------|
| `g()` value default, free-func generator | 11 | **NaN** |
| `new C().gen()` ref default, instance generator | x123 | **6** |
| `ag()` value default, async generator | 11 | **NaN** |
| any of the above **with the argument supplied** (`g(100)`, `gen("Y")`) | correct | correct ✓ |

Compiled `g()` yields `NaN` because the omitted `acc` reaches the closure as the `$Undefined` sentinel and `$Undefined + 1` is `NaN`; the ref case yields `6` because the omitted `s` is treated as `0` (`0 + 1 + 2 + 3`) — in both, the default (`5` / `"x"`) never reached the closure's copy of the parameter.

## Root cause

[...]

Issue #790, opened 2026-06-16
Compiled: base-typed call to a derived override that adds trailing optional params dispatches to the base method (override ignored)


nickna:
## Summary

In **compiled** mode, when a derived class overrides a base method by **adding one or more trailing optional / default parameters** (a valid TypeScript override — the signature stays assignable), a call made through a **base-typed reference** dispatches to the **base** method instead of the derived override. The derived method's added default therefore never fires, and the override is effectively invisible to base-typed call sites. The interpreter is correct.

This is **not** a default-parameter-firing bug: it reproduces with a **reference-type** default too (so the default mechanism itself is fine), and a **direct** call on the derived static type works. The root cause is virtual dispatch to a wider-arity override.

## Repro

```ts
// value-type added default
class Base { m(x: number): number { return x; } }
class Derived extends Base { m(x: number, y: number = 100): number { return x + y; } }

const d = new Derived();
console.log(d.m(3));          // A  interp: 103   compiled: 103   (direct call — OK)

const b: Base = new Derived();
console.log(b.m(3));          // B  interp: 103   compiled: 3     (BUG — dispatches to Base.m)

// reference-type added default — shows it is NOT default-firing-specific
class B2 { g(x: number): string { return "x" + x; } }
class D2 extends B2 { g(x: number, s: string = "Z"): string { return "x" + x + s; } }
const b2: B2 = new D2();
console.log(b2.g(3));         // C  interp: x3Z   compiled: x3    (BUG — dispatches to B2.g)
```

| case | scenario | interp | compiled |
|------|----------|--------|----------|
| A | direct `Derived.m(3)` (static type `Derived`) | 103 | 103 ✓ |
| B | base-typed `b.m(3)`, value-type added default | 103 | **3** ✗ |
| C | base-typed `b2.g(3)`, reference-type added default | x3Z | **x3** ✗ |

Compiled case B returns `3` and case C returns `"x3"` — exactly the results of invoking `Base.m`/`B2.g`, confirming the base-typed call binds to the base method and never reaches the override.

## Root cause

`Derived.m(x, y = 100)` is a valid TS override of `Base.m(x)` because `(x: number, y?: number) => number` is assignable to `(x: number) => number`. But its CLR arity differs from the base (2 params vs 1), so the IL emitter defines it as a **new (`newslot`) method** rather than overriding the base's vtable slot. A statically-`Base`-typed call resolves to `Base.m(double)` and dispatches there.

[...]

Issue #739, opened 2026-06-16
Compiled: direct call to an instance/static method or constructor pads an omitted optional (no-default) param with null, not the `undefined` sentinel


nickna:
## Summary

In **compiled** mode, calling a class **instance or static method** (or a constructor) **directly** with fewer arguments than parameters pads a missing trailing **optional, no-default** parameter (`x?: T`) with CLR `null` instead of the `undefined` sentinel. Observable through `typeof`, `=== undefined`, `=== null`. The interpreter is correct, and free functions are correct.

This is the direct-call analogue of #640 (which fixed the **value-call** / `$TSFunction.Invoke` path) and #703 (class methods invoked as values). The direct-call path for class methods was not covered.

## Repro

```ts
class C {
  m(x?: any): string { return typeof x; }
  static s(x?: any): string { return typeof x; }
}
console.log(new C().m());   // interp: "undefined"   compiled: "object"  (BUG)
console.log(C.s());         // interp: "undefined"   compiled: "object"  (BUG)
```

`typeof null === "object"`, so the missing optional reads as `null`. Free functions are fine (`function h(x?: any) { return typeof x; } h()` → `"undefined"` in both modes). Class **expressions** are also fine (they use all-`object` param slots).

## Cause / fix sketch

The direct method-call emitter pads missing trailing args with the slot's CLR default (null for object slots) rather than the `$Undefined` singleton. For object-typed params it should pad with the sentinel (cf. the private-method call padding added in #705, `ILEmitter.Properties.Private.cs` `EmitPrivateCallArgsPadded`, which pads with `$Undefined`). Mirror that for the instance/static method and constructor direct-call sites. Note this is distinct from default-parameter firing — these params have no default; the value should simply be `undefined`.

## Discovered
While implementing #703/#705.

nickna:
Verified resolved on main (e8f39905): an omitted trailing optional (no-default) param reads undefined (typeof => undefined) for instance and static methods in compiled mode, matching the interpreter. Implemented in #787.

Issue #738, opened 2026-06-16
Compiled --ref-asm: value-call ignores a value-type default parameter


nickna:
## Summary

Follow-up to #705 (case b). In **`--compile --ref-asm`** (reference-assembly emit) mode, calling a function expression / arrow **as a value** does not apply a value-type parameter default for an omitted argument. Plain `--compile` and the interpreter are correct.

## Repro

```ts
const f = function (x: number, y: number = 3) { return x + y; };
console.log(f(4));
```

| mode | result |
|------|--------|
| interpreter | `7` |
| `--compile` | `7` |
| `--compile --ref-asm` | `3`  ← BUG (default `y = 3` not applied, `y` is 0) |

## Notes

- Pre-existing and independent of #705's fix: #646 already widened arrow/function-expression defaulted params to an `object` slot, and plain `--compile` applies the default via the entry prologue. The `--ref-asm` value-call path does not. (Before #646 this same case produced an invalid-IL `StackUnexpected` error, so `--ref-asm` was already broken here.)
- Audit the `--ref-asm` value-call dispatch for widened `object` params — the omitted slot is not being padded with the `$Undefined` sentinel (or the entry prologue is not reached) on that path.

## Discovered
While implementing #705.

nickna:
Verified resolved on main (e8f39905): compile --ref-asm value-call applies the value-type default (f(4)=>7), matching --compile and the interpreter. Implemented in #787 (the $ExpectsThis marker attribute).

PR #794, opened 2026-06-16
Fix un-buildable main: ParameterInfo[]/Type[] mismatch in generator instance stubs (re-enables #723/#737/#738/#739)

### LINKED ISSUE #720: Compiled mode: async/generator private methods (`async #m`, `*#m`) emit invalid IL / unsupported
## Summary

In **compiled mode**, an **async** or **generator** ES2022 private method is not emitted through a state machine — `EmitPrivateMethodBody` (`Compilation/ILCompiler.Classes.Methods.cs`) emits the body linearly into the `__private_<name>` method regardless of `IsAsync`/`IsGenerator`. The interpreter handles both correctly.

Discovered while completing #696 (private-method optional params). Independent of #696 — reproduces with **required** params that parsed before #696.

## Repro — async private method

```ts
class Q {
  async #p(x: number): Promise<number> { return x; }
  go(): Promise<number> { return this.#p(5); }
}
new Q().go().then(v => console.log(v));
```

`--compile … --verify`:
```
[IL Error] Q.__private_p: StackUnexpected {Offset=1, Found=ref 'object', Expected=ref 'Task`1<object>'}
```
The method is declared to return `Task<object>` (see `ILCompiler.Classes.cs` private-method definition: `returnType = method.IsAsync ? _types.TaskOfObject : typeof(object)`) but the linear body `return x` leaves a bare `object` on the stack. It happens to run (JIT lenient + compiled `await` of a non-task returns the value), but the IL is invalid.

## Repro — generator private method

```ts
class G {
  *#p(x: number): Generator<number> { yield x; yield x + 1; }
  go(): number { let s = 0; for (const v of this.#p(5)) s += v; return s; }
}
console.log(new G().go());
```
`--compile`:
```
Compile Error: Yield not supported in this context
```

## Expected

Async/generator private methods compile through the same state-machine path as their public counterparts (`AsyncStateMachineBuilder` / generator lowering), so `__private_p` returns a real `Task<object>` / `IEnumerable<object>` and `yield` is supported.

## Fix sketch

`EmitPrivateMethodBody` needs to branch on `method.IsAsync`/`method.IsGenerator` and drive the corresponding state-machine builder instead of emitting the body inline, mirroring how public class methods are emitted. Parameter defaults (the prologue added in #696) should then move into the state-machine entry prologue (as `AsyncMoveNextEmitter` already does for public methods).

### LINKED ISSUE #723: Compiled: class methods (instance + static) ignore default parameter values
## Summary

In **compiled** mode, **class methods** — both instance and static — ignore default parameter values entirely. A missing argument is left at its CLR zero value (`0` for `number`, `null` for reference types) instead of the parameter's default. The **interpreter** handles all of these correctly. Top-level **function declarations** and **arrow / function expressions** apply defaults correctly in compiled mode (see #646); only class methods are affected.

## Repro

```ts
class C {
  m(a: number, b: number = 2) { return a + b; }
  s2(a: string, b: string = "z") { return a + b; }
  static st(a: number, b: number = 2) { return a + b; }
}
console.log(new C().m(5));     // interpreted: 7    compiled: 5      (b dropped → 0)
console.log(new C().s2("x"));  // interpreted: xz   compiled: xnull  (b dropped → null)
console.log(C.st(5));          // interpreted: 7    compiled: 5
```

When every argument is supplied the methods behave correctly (e.g. `new C().m(5, 10)` ⇒ `15` in both modes); only the omitted-argument (default) path is broken.

## Root cause

`EmitMethod` (`Compilation/ILCompiler.Classes.Methods.cs`) defines the method's parameters but never applies defaults: unlike the arrow path it calls neither `ParameterTypeResolver.WidenDefaultedParamsToObject` nor `ILEmitter.EmitDefaultParameters`, and unlike the top-level-function path it generates no `OverloadGenerator` lower-arity overloads. So a call supplying fewer arguments than parameters pads the missing slots with the CLR default and the body observes `0`/`null`.

## Why this is filed separately from #698

#698 (a default that references an earlier parameter) is now fixed for arrows / function-expressions / function-declarations in both modes, and for methods in the **interpreter**. The compiled-method earlier-param case (`m(a, b = a * 2)` ⇒ `5` instead of `15`) is a strict subset of *this* broader gap — compiled methods drop **all** defaults, not just earlier-param ones — so it cannot be closed without fixing this first.

Fixing it needs care around **virtual dispatch / override-signature matching**. The arrow-style fix (widen defaulted params to `object`) changes a method's CLR signature and would break override / interface-implementation matching against a base or subclass whose same-named method has no defaults (e.g. `class Base { m(a, b = 2) {} }` vs `class Derived extends Base { m(a, b) {} }` — both must stay signature-compatible). The safer route is likely method-level arity overloads analogous to `OverloadGenerator` for free functions, with method dispatch selecting the right arity.

(Class *expression* methods in `Compilation/ILCompiler.Classes.ClassExpressions.cs` DO call `EmitDefaultParameters` and so may behave differently — not verified here. Note that path applies defaults via the null/`$Undefined` check, which cannot observe a value-type `double`/`bool` slot, cf. #705.)

## Discovered

While implementing #698. Interpreter parity confirmed correct for all the above.

nickna:
## What this fixes

`origin/main` currently **does not compile**:

```
Compilation/ILCompiler.AsyncGenerators.cs(376,107): error CS1503: Argument 6: cannot convert from 'System.Reflection.ParameterInfo[]' to 'System.Type[]?'
Compilation/ILCompiler.Generators.cs(557,107): error CS1503: Argument 6: cannot convert from 'System.Reflection.ParameterInfo[]' to 'System.Type[]?'
```

This is a **semantic merge conflict** between two independently-correct PRs that were merged without being rebased onto each other:

- **#769** (`e64f3adb`, compiled closures #724/#725) added `EmitGeneratorFunctionDCInit(…, Type[]? paramTypes)` and called it from the generator / async-generator **instance** stub methods, passing their `paramTypes` local (then a `Type[]`).
- **#786** (`b95b6945`, "Fix #720: async/generator private methods emit invalid IL") changed that same `paramTypes` local from the AST-resolved `Type[]` to `methodBuilder.GetParameters()` (a `ParameterInfo[]`) — deliberately, so boxing is decided from the method's **actual IL signature** (a private method's params are all `object` slots; boxing an AST-resolved `double` into an `object` slot is a `StackUnexpected`).

Each compiled in isolation; together a `ParameterInfo[]` flows into a `Type[]?` parameter → CS1503. (Neither side is #787 — that is simply what surfaced this when I went to verify it.)

## The fix

Convert at the two call sites with `.Select(p => p.ParameterType).ToArray()` — the dominant idiom in this codebase for `GetParameters()` → `Type[]` (`ILCompiler.Classes.Methods.cs:1250`, `Static.cs:438`, `Constructors.cs:363`). This preserves #720's "box from the actual IL signature" intent: for a private/all-object or #737-widened-to-`object` slot, `ParameterType` is `object`, so `IsValueType` is false and nothing is spuriously boxed — exactly matching the field-seeding loop directly above each call.

```diff
-            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1, paramTypes);
+            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1,
+                paramTypes.Select(p => p.ParameterType).ToArray());
```

(I considered widening the helper to `ParameterInfo[]?` instead, but it genuinely operates on `Type` — it boxes with the element and checks `.IsValueType` — and two of its four callers pass `null`; the call-site conversion is the smaller, more idiomatic change.)

## Why this matters / relation to #723 #737 #738 #739

The compiled default/optional-parameter cluster (#705/#723/#737/#738/#739) was implemented and merged in #787, but the later merge above left `main` un-buildable, so those features could not actually run. With the build repaired I verified all four issues' repros, compiled output matching the interpreter:

- **#723** instance/static method defaults — `new C().m(5)`→`7`, `C.st(5)`→`7`, string default→`xz` ✓
- **#737** value-type defaults on instance/static + generator methods, incl. the same-arity override-safety case ✓
- **#738** `--compile --ref-asm` value-call no longer shifts args (`f(4)`→`7`) ✓
- **#739** omitted optional reads `undefined` (`typeof`→`"undefined"`) ✓

Existing coverage `Generator_InstanceMethodCallbackMutatesCapturedParameter` — an instance generator method whose arrow mutates a captured **value-type** parameter, the exact path repaired here — exercises the boxing branch and passes. Full unit suite green (13352 passed, 0 failed, 1 skipped).

## Follow-ups filed (pre-existing gaps surfaced while verifying)

- **#790** — a base-typed call to a derived override that *adds* trailing optional params dispatches to the base method (a different-arity override is not a CLR vtable override). Affects value- and reference-type defaults and no-default optionals; distinct from the same-arity override group #787 handles.
- **#792** — a defaulted generator/async-generator parameter that is *also* captured by a nested arrow ignores its default on omitted args (the default prologue writes the state-machine field, but a captured param's live storage is the function-DC field). Intersection of #737 and #724/#725; reproduces for free-function generators too, so independent of this build fix.

No new tests added: the break is compile-time (caught by any build) and its runtime path is already covered; the two follow-ups are failing cases tracked as issues.

Fixes #723, #737, #738, #739.

### DIFF
--- a/Compilation/ILCompiler.AsyncGenerators.cs
+++ b/Compilation/ILCompiler.AsyncGenerators.cs
@@ -371,9 +371,12 @@ public partial class ILCompiler
 
         // #725: instantiate the function display class and seed any captured-and-mutated parameter
         // into it (instance methods carry 'this' at arg 0, so user params start at arg 1). The stub
-        // params are typed, so value types are boxed before the store. No-op when no function DC.
+        // params are typed, so value types are boxed before the store. No-op when no function DC. Pass
+        // the CLR boxing types from the method's ACTUAL IL signature (same source as the field-seeding
+        // loop above), so widened/all-object slots are not spuriously boxed.
         if (funcDCKey != null)
-            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1, paramTypes);
+            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1,
+                paramTypes.Select(p => p.ParameterType).ToArray());
 
         // Return the state machine (which implements IAsyncEnumerable<object>)
         il.Emit(OpCodes.Ret);
--- a/Compilation/ILCompiler.Generators.cs
+++ b/Compilation/ILCompiler.Generators.cs
@@ -552,9 +552,12 @@ public partial class ILCompiler
         // into it so an arrow that writes that parameter shares the generator's storage. Instance
         // methods carry 'this' at arg 0, so user params start at arg 1 (paramOffset: 1). The stub
         // params are typed, so EmitGeneratorFunctionDCInit boxes value types before the store. No-op
-        // when the method has no function DC.
+        // when the method has no function DC. Pass the CLR boxing types from the method's ACTUAL IL
+        // signature (same source as the field-seeding loop above), so widened/all-object slots are not
+        // spuriously boxed.
         if (funcDCKey != null)
-            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1, paramTypes);
+            EmitGeneratorFunctionDCInit(il, smBuilder.FunctionDCField, method, funcDCKey, paramOffset: 1,
+                paramTypes.Select(p => p.ParameterType).ToArray());
 
         // Captured outer-scope variables are NOT copied into the state machine (#541): see the
         // free-function stub above. MoveNext reads/writes them live from their backing storage,


nickna:
Closing as **superseded** — a concurrent session fixed the same build break first.

This PR repaired an un-buildable `main` (CS1503 ×2: `ParameterInfo[]` → `Type[]?` in the generator/async-generator instance stubs, from the #769×#786 merge seam) by converting at the two call sites. While it was open, `main` advanced to `e8f39905`, where commit `35640416` ("Fix broken build: EmitGeneratorFunctionDCInit param type (ParameterInfo[])") already fixed the identical break via the equivalent **opposite** approach — widening the helper's parameter to `ParameterInfo[]?` so the stub's `methodBuilder.GetParameters()` flows in directly. Both are correct and box from the actual IL signature; theirs landed first.

CI here is red only because the PR is built **merged with the now-fixed main**, which produces the reverse mismatch (`Type[]` → `ParameterInfo[]?`). There is no longer any code change to make.

Verification still stands and was re-run against current `main` (`e8f39905`):

- **#723 / #737 / #738 (`--ref-asm`) / #739** — all confirmed working, compiled output matching the interpreter.
- Filed **#790** (a base-typed call to a derived override that *adds* trailing optional params dispatches to the base method) and **#792** (a defaulted generator/async-generator parameter that is *also* captured by a nested arrow ignores its default on omitted args) as pre-existing gaps surfaced while verifying — both still reproduce on `e8f39905`.

Issue #886, opened 2026-06-23
Compiled: assigning to RegExp.lastIndex emits unverifiable IL (StackUnexpected: ref float64 vs Double)


nickna:
## Summary

In **compiled** mode, assigning to `RegExp.prototype.lastIndex` emits IL that fails ILVerify. The program still runs and produces the correct result (the CLR is lenient here), but `--compile … --verify` reports a stack-type error, so the emitted assembly is not formally verifiable.

## Minimal repro

```ts
const r = /a/;
r.lastIndex = 5;
```

```
dotnet run -- --compile repro.ts -o repro.dll --verify
```

```
IL verification found 1 error(s):
  [IL Error] $Program.Main: StackUnexpected {Offset=57, Found=ref 'float64', Expected=Double} - Unexpected type on the stack.
```

The assignment is the trigger — a read alone is fine:

| Source | `--verify` | runs |
|---|---|---|
| `console.log(/a/.lastIndex);` | passes | `0` |
| `const r = /a/; console.log(r.lastIndex);` | passes | `0` |
| `const r = /a/; r.lastIndex = 5;` | **StackUnexpected** | (ok) |
| `const r = /a/; r.lastIndex = 5; console.log(r.lastIndex);` | **StackUnexpected** | `5` (matches interpreter) |

## Expected vs actual

- **Expected:** `r.lastIndex = 5` emits verifiable IL.
- **Actual:** ILVerify finds `Found=ref 'float64', Expected=Double` — a reference/boxed `float64` reaches `RegExpSetLastIndex(object, double)` where the second parameter wants a native `float64`. The program runs correctly regardless.

## Likely location

`Compilation/ILEmitter.Properties.cs`, the `RegExp.lastIndex` setter fast path (~lines 643-661). It skips the box -> unbox-through-`Convert` step when `_stackType == StackType.Double`:

```csharp
EmitExpression(s.Value);
if (_stackType != StackType.Double)
{
    EmitBoxIfNeeded(s.Value);
    EmitUnboxToDouble();
}
// ... Call RegExpSetLastIndex(object, double)
```

The symptom suggests `_stackType` reads as `Double` for this value shape while the value actually on the stack is a boxed/by-ref `float64`, so the box->unbox normalization is skipped and `RegExpSetLastIndex`'s native-`double` parameter receives a reference. (Hypothesis — not root-caused.)

## Notes

- **Verification-only:** runtime behavior is correct in both interpreter and compiled modes; only `--verify` fails.
[...]

PR #887, opened 2026-06-23
Regex perf: close the compiled-mode gap to Node (EnumerateMatches + Compiled + construction cache + literal hoisting)

nickna:
## Regex performance: close the compiled-mode gap to Node

Investigation started from the cross-runtime benchmark, where compiled regex was **3.2× slower than Node** at the meaningful sizes. A/B measurement showed the .NET `Regex` *engine* was not the bottleneck — result materialization and per-evaluation construction were. This PR lands four semantics-preserving levers, each measured.

### What's in here

| # | Lever | Effect |
|---|---|---|
| 1 | **Alloc-light match path** — `String.prototype.match(/g)` iterates `Regex.EnumerateMatches` (span-based `ValueMatch`) instead of `Regex.Matches()`, dropping the per-match `Match`/`Group` allocation | match path ~2.3× |
| 2 | **`RegexOptions.Compiled`** on the cached engine (was running interpreted; amortized by the existing `(pattern, options)` compile cache) | ~1.3× on raw matching |
| 3 | **Wrapper-construction template cache** — validate/normalize/scan/compile once per `(pattern, flags)` instead of every construction | 1.3–4.1× per construction |
| 4 | **Compiled regex-literal hoisting** — a constant `/lit/` in a safe consuming position becomes one static `$RegExp` per site (V8-style, lazy `ldsfld`/init) instead of `new $RegExp` per evaluation | literal-in-loop 2.1× |

Levers 1+2 apply to both the interpreter and the compiled `$RegExp` emitter. Lever 3 is the interpreter runtime type. Lever 4 is compiled-only.

### Measured (compiled, Windows x64, .NET preview)

- `regex.ts` bulk match, N=10000: **6.85 ms → 3.44 ms** — compiled-vs-Node **3.20× → 1.61×**.
- Literal-in-a-loop `/^\w+@\w+\.\w+$/.test(...)`, N=100k: **10.19 ms → 4.86 ms (2.1×)**, hitting the hand-hoisted floor (4.76 ms).

### Safety notes for review

- **Levers 1–3 are semantics-preserving by construction**: `Compiled` changes *how* not *what*; `EnumerateMatches` returns the same matches as `Matches`; the template cache runs identical validation, just memoized (invalid patterns still throw `SyntaxError` on every construction, validation ordering preserved).
- **Lever 4 safety gate** (`RegexLiteralHoistAnalyzer`): hoist `.test`/`.exec` receivers **only without `g`/`y`** (they advance `lastIndex` across evaluations); hoist `match`/`matchAll`/`replace`/`replaceAll`/`search`/`split` args always (those scan from 0 and never read instance `lastIndex`); exclude anything that escapes. The set/field map are **reference-keyed** (`Expr.RegexLiteral` is a record — value-keying would let an escaping `const r = /a/` wrongly share a hoisted `/a/.test()`). Wired into both `Compile` and `CompileModules`.

### Tests

- +10 dual-mode hoisting regression tests (`RegExpLiteralHoistingTests`): `g`/`y` exclusion, escaping mutation isolation, identity, all six string consumers — interpreter as oracle, compiled must match.
- Full unit suite **14,000 green**; regex / IL-verification / standalone-DLL / closure / module suites green; `--verify` clean on all hoisted shapes.
- **Test262 run is pending** (regex is conformance-sensitive — covers `lastIndex`/identity edges that gate lever 4); results to follow before merge.

### Deferred / out of scope (documented)

- The remaining ~1.6× on bulk and ~1.9× on literal-loops vs Node is the per-`.test()` dispatch + `Regex.Replace` engine throughput — a fundamental .NET-vs-Irregexp gap, not closeable by these techniques. See `docs/plans/regex-literal-hoisting-scope.md`.
- An ECMAScript-mode-drop lever (~1.2×) is deferred pending Test262 — it changes matching semantics.

### Unrelated find

Filed #886 — `console.log`/assignment to `RegExp.lastIndex` emits unverifiable IL on `main` (pre-existing, not touched here).

### DIFF
--- a/Compilation/EmittedRuntime.cs
+++ b/Compilation/EmittedRuntime.cs
@@ -39,6 +39,18 @@ public class EmittedRuntime
     /// </summary>
     public void RequireSharpTSRuntime(string reason) => RequiredSharpTSRuntimeReasons.Add(reason);
 
+    /// <summary>
+    /// Per-site hoisted regex-literal static fields (compiled-mode optimization),
+    /// keyed by <c>Expr.RegexLiteral</c> node <em>identity</em> — see
+    /// <see cref="RegexLiteralHoistAnalyzer"/>. Null when nothing was hoisted.
+    /// <c>EmitRegexLiteral</c> loads the field (constructing the <c>$RegExp</c>
+    /// once, lazily) instead of allocating a fresh instance per evaluation. Lives
+    /// on the shared <see cref="EmittedRuntime"/> so it is reachable from every
+    /// emission context (functions, arrows, state machines) with no extra
+    /// threading.
+    /// </summary>
+    public Dictionary<SharpTS.Parsing.Expr.RegexLiteral, FieldBuilder>? RegexHoistFields { get; set; }
+
     // The emitted $Undefined singleton class
     public Type UndefinedType { get; set; } = null!;
     public FieldInfo UndefinedInstance { get; set; } = null!;
--- a/Compilation/ILCompiler.cs
+++ b/Compilation/ILCompiler.cs
@@ -430,6 +430,7 @@ public partial class ILCompiler
         ObjectLocalPromotionAnalyzer.Analyze(statements, _typeMap, _closures.Analyzer);
         Phase3_CreateProgramType();
         DefineObjectShapeTypes();
+        DefineHoistedRegexFields(statements);
         PreScanBuiltInModuleImports(statements);
         Phase4_DefineDeclarations(statements);
         Phase5_CollectArrowFunctions(statements);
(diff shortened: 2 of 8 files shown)


nickna:
## Test262 conformance results

Ran the committed Test262 subset (11,384 tests across 13 folders, including `built-ins/RegExp` and `built-ins/String` — the surface this PR touches), one mode at a time.

### Compiled mode — clean pass ✅

`CompiledBaseline` **passes with zero drift** against `baselines/compiled.txt` (no regressions, no new passes, no new/removed entries; 15m30s). This is the mode where the literal hoisting and the `$RegExp` emitter changes live, so an exact baseline match is the result that matters most here.

### Interpreted mode — pre-existing flaky drift, not from this PR

`InterpretedBaseline` reports drift, but it is **pre-existing baseline staleness/flakiness, not caused by this branch**:

| | Regressions | New passes | Where |
|---|---:|---:|---|
| `origin/main` (no regex changes) | 12 | 7 | `Array` / `Number/toString` / `Object/defineProperty` / `String`-ctor |
| this PR | 10 | 5 | same families |

- **`main` itself fails the gate** (12/7) — the interpreted baseline is stale relative to current `main`.
- The drift sets **differ between runs** (e.g. `main` has `Array/prototype/map` + `Array/prototype/indexOf`; this branch a different `Object/defineProperty` test) → nondeterministic harness noise (timeout/GC/ordering-sensitive), not a deterministic code change.
- **No `RegExp` or regex-driven `String` test appears in either drift set.** This PR only touches regex code, which cannot affect `Array.join` / `Number.toString` / `Object.defineProperty`. This branch's drift is in fact *smaller* than `main`'s.

### Conclusion

No conformance regression. Compiled mode (where the optimization lives) matches the baseline exactly; the interpreted "failure" is pre-existing harness flakiness that affects `main` equally.

<sub>Harness notes for reproducing: build `SharpTS.Test262.Worker` in Release first (else it falls back to crash-prone in-process execution), and run one mode at a time (`--filter "FullyQualifiedName~Test262InterpretedTests"` / `~Test262CompiledTests`) — the two baseline facts otherwise run concurrently and exhaust the host.</sub>

PR #932, opened 2026-06-25
fix(interp): dispatch Error/instance toString in string coercion (#922)

### LINKED ISSUE #922: Interpreter: String(error)/""+error/template literal skip Error.toString() (give "RangeError instance"/"[object RangeError]")
## Summary

In the **interpreter**, string-coercing an `Error` object via `String(e)`, `"" + e`, or a template literal `` `${e}` `` does not invoke the error's `toString()` — it falls back to a default representation. Direct `e.toString()` is correct. **Compiled mode is correct** for all forms and matches Node.

## Repro

```ts
const e = new RangeError("real range error");
console.log(String(e));     // expected: RangeError: real range error
console.log("" + e);        // expected: RangeError: real range error
console.log(`${e}`);        // expected: RangeError: real range error
console.log(e.toString());  // expected: RangeError: real range error
```

## Actual vs expected

| Expression | Node / Compiled | Interpreter |
|---|---|---|
| `String(e)` | `RangeError: real range error` | `RangeError instance` ✗ |
| `"" + e` | `RangeError: real range error` | `[object RangeError]` ✗ |
| `` `${e}` `` | `RangeError: real range error` | `[object RangeError]` ✗ |
| `e.toString()` | `RangeError: real range error` | `RangeError: real range error` ✓ |

## Root cause (suspected)

The interpreter's `ToPrimitive` / string-coercion for a `SharpTSError` (and likely error subclasses) doesn't dispatch to the object's `toString()`. Note the two wrong outputs differ — `String()` yields `"RangeError instance"` while `+`/template yield `"[object RangeError]"` — so more than one coercion path skips `toString()`. This mirrors the array→string coercion bug fixed earlier in `ToPrimitive` (`'' + [1,2,3]`); errors need the same treatment.

## Notes

Found incidentally while fixing the cross-boundary guest-error identity gap (#694 follow-up, PR #920). Not a regression from that PR — pre-existing, interpreter-only.

nickna:
## Summary

Fixes #922. In the **interpreter**, string-coercing a class instance skipped its `toString()`:

| Expression | Before (interp) | After (interp + compiled + Node) |
|---|---|---|
| `String(e)` | `RangeError instance` | `RangeError: real range error` |
| `"" + e` | `[object RangeError]` | `RangeError: real range error` |
| `` `${e}` `` | `[object RangeError]` | `RangeError: real range error` |
| `e.toString()` | ✓ already correct | ✓ |

Two distinct paths skipped `toString` (hence the two different wrong outputs):
- `Stringify` (`Interpreter.Operators.cs`) hard-coded `"[object ClassName]"` for `SharpTSInstance` — served `"" + e` and `` `${e}` ``.
- `SharpTSStringNamespace.Call` routed only `SharpTSObject` through `ToStringForStringCall`; instances fell through to `SharpTSInstance.ToString()` (`"ClassName instance"`) — served `String(e)`.

Fix: a `TryStringifyInstanceToString` helper dispatches an instance's `toString` (own-field shadow first, else up the class chain via `FindMethod`/`BindMethod`), wired into `Stringify`; and `SharpTSInstance` now routes through `ToStringForStringCall`. Generalizes to any user class with a custom `toString`, matching compiled mode and Node.

## Follow-up (same root cause): array element coercion

`join` / `Array.prototype.toString` / `String(arr)` / `"" + arr` / `` `${arr}` `` used a debug stringifier that skipped `toString`, mis-rendered nested arrays, and inspect-formatted objects — in **both** modes:

| Case | Before (interp / compiled) | After (both) |
|---|---|---|
| `[err].join()` / `"" + [err]` | `Error instance` / `Error: m` | `Error: m` |
| `[[1,2],[3]].join("-")` | `[1, 2]-[3]` | `1,2-3` |
| `[{a:1}].join()` | `{ a: 1 }` | `[object Object]` |

- **Interpreter** (`ArrayBuiltIns`): new `CoerceToJsString` does ECMA ToString per element/separator/default-sort key (nested array → recursive join; instance/Error → `toString`; plain object → `[object Object]`; wrappers unwrapped).
- **Compiled** (`EmitArrayJoin`, `ToJsString` array branch): call `ToJsString` per element instead of the debug `Stringify` — the latter's comment already intended recursion.

## Tests
- New shared tests (`ArrayMethodTests`, `ErrorTests`): errors / nested arrays / plain objects across both modes; interpreted-only for user-class `toString`.
- Corrected `Spread_Map_CreatesArrayOfTuples`, which **asserted the pre-fix debug output** (`[x, 10]`) — now `x,10` to match Node, with a mode-dependent expectation documenting the compiled `object[]`-tuple gap below.
- Full suite green apart from the pre-existing stale **untracked** `SharpTS.Tests/Test262/` baseline (identical drift, no new regressions).

## Discovered, filed separately (pre-existing compiled-only gaps)
1. Compiled doesn't dispatch user-class (non-Error) `toString` in any string coercion (`String(new Pt())` → `"Pt"`). #922's "compiled is correct" only held because it tested `Error` (specially handled).
2. Compiled `object[]` (Map/Set/generator spread) tuples hit `ToJsString`'s debug fallback.

### DIFF
--- a/Compilation/RuntimeEmitter.Arrays.Search.cs
+++ b/Compilation/RuntimeEmitter.Arrays.Search.cs
@@ -1283,7 +1283,7 @@ public partial class RuntimeEmitter
         il.Emit(OpCodes.Br, setSepLabel);
         il.MarkLabel(hasSep);
         il.Emit(OpCodes.Ldarg_1);
-        il.Emit(OpCodes.Call, runtime.Stringify);
+        il.Emit(OpCodes.Call, runtime.ToJsString);
         il.MarkLabel(setSepLabel);
         il.Emit(OpCodes.Stloc, sepLocal);
 
@@ -1338,10 +1338,13 @@ public partial class RuntimeEmitter
         il.Emit(OpCodes.Isinst, runtime.UndefinedType);
         il.Emit(OpCodes.Brtrue, skipAppend);
 
-        // sb.Append(Stringify(elem))
+        // sb.Append(ToJsString(elem)) — ECMA-262 23.1.3.16 step 7 ToString per element:
+        // a nested array joins recursively and an object / class instance (incl. Error)
+        // dispatches its toString, rather than the console/debug Stringify form
+        // ("[1, 2]" / "{ a: 1 }" / "ClassName") (#922 follow-up).
         il.Emit(OpCodes.Ldloc, sbLocal);
         il.Emit(OpCodes.Ldloc, elemLocal);
-        il.Emit(OpCodes.Call, runtime.Stringify);
+        il.Emit(OpCodes.Call, runtime.ToJsString);
         il.Emit(OpCodes.Callvirt, _types.GetMethod(_types.StringBuilder, "Append", _types.String));
         il.Emit(OpCodes.Pop);
 
--- a/Compilation/RuntimeEmitter.CoreUtilities.cs
+++ b/Compilation/RuntimeEmitter.CoreUtilities.cs
@@ -1934,7 +1934,7 @@ public partial class RuntimeEmitter
         il.Emit(OpCodes.Brtrue, skipAppend);
         il.Emit(OpCodes.Ldloc, sbJoinLocal);
         il.Emit(OpCodes.Ldloc, valLocalJ);
-        il.Emit(OpCodes.Call, runtime.Stringify);
+        il.Emit(OpCodes.Call, runtime.ToJsString);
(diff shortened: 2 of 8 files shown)


nickna:
Filed the two discovered compiled-only gaps as separate issues: #933 (user-class `toString` not dispatched in compiled string coercion) and #934 (compiled `object[]` Map/Set/generator spread tuples render in debug form).

Issue #928, opened 2026-06-25
Typed-array perf: native-int arithmetic path (close int32 ~3.7x Node) + double[]/Span+SIMD stencil floor


nickna:
## Summary

Follow-up perf levers deferred from the typed-array compiled-perf arc (**PR #892**), which brought Float64/compound typed arrays to ~1.14× Node but left integer-array kernels at ~3.7× Node.

## 1. Native-int arithmetic path for integer typed arrays

`Int32Array`/`Uint32Array`/etc. kernels measure **~3.7× Node** (int32-kernel 76→9.5ms vs Node). The residual is per-element `int↔double` conversion plus a native-int store that V8 has and SharpTS lacks — SharpTS reinterprets everything through `double`. Adding a native-int arithmetic + store path for integer-element arrays should close most of that 3.7×.

## 2. `double[]`/Span + SIMD for the stencil-read floor

The residual 1M stencil-read floor on Float64 could be lowered with a `double[]`/`Span<double>` backing + SIMD. **Caveat:** this is the only interop-*risky* option — every typed array currently backs onto a shared `byte[]` (matching V8/Bun), which preserves `ArrayBuffer`/`DataView`/`Atomics` aliasing. Changing the backing representation must not break that aliasing, so this needs careful design (or a parallel representation gated to non-escaping cases).

## Context

Representation today: every typed array (interp + compiled) backs onto `byte[]`; element access reinterprets via `Unsafe.Read/WriteUnaligned`. See `Runtime/Types/SharpTSTypedArray.cs`, `Compilation/RuntimeEmitter.TSTypedArray.cs`. Part of epic #856.

nickna:
## Closing — the practical wins shipped; both named levers are blocked by the constraints this issue itself states

Re-investigated with split-isolation benchmarks (compiled vs Node v25 / Bun 1.3 @1M). Summary of where this lands:

### The two levers named here are dead ends (by the stated non-negotiables)
[...]

Issue #882, opened 2026-06-22
Test262: 62 deterministic compiled-mode regressions vs stale baseline (14 compiled-only codegen bugs)


nickna:
## Summary

A clean run of the **compiled** Test262 baseline (`Test262CompiledTests.CompiledBaseline`) against the committed `baselines/compiled.txt` shows **62 regressions** and **108 new passes** (net **+46**; suite still improved overall). All 62 regressions are **deterministic** — the 20 sampled in the run log reproduced 20/20, and the full set was reconstructed and independently re-verified (882 candidate paths re-run → exactly 62, zero flake).

The root context: **`baselines/compiled.txt` is 187 commits stale** — last regenerated at `0494a763` (#693, 2026-06-15), while `baselines/interpreted.txt` is current as of `b9d3aa3f` (#802, 2026-06-17). So this drift is the *accumulated* compiled-mode delta across 187 commits, never recorded.

Cross-referencing each regressed path against the (current) interpreted baseline splits the 62 cleanly:

| Class | Count | Meaning |
|---|---:|---|
| **Compiled-only** (interp still `Pass`) | **14** | Genuine compiled/codegen regressions — interpreter gets these right, compiled does not. **Priority.** |
| Shared (interp baseline already `Fail`/`RuntimeError`) | 48 | Behavior changes already accepted on the interpreted side at #802; the stale compiled baseline just never recorded them. |

Buckets: 37 `RuntimeError` + 25 `Fail`.

## Reproduce

```bash
dotnet build SharpTS.Test262.Worker/SharpTS.Test262.Worker.csproj -c Release   # worker is NOT rebuilt by dotnet test
dotnet test SharpTS.Test262/SharpTS.Test262.csproj -c Release \
  --filter "FullyQualifiedName~CompiledBaseline"        # ~17 min; run backgrounded
# Regenerate (only after the 14 compiled-only are triaged — don't bury them):
#   SHARPTS_TEST262_UPDATE_BASELINE=1 dotnet test ... --filter "FullyQualifiedName~CompiledBaseline"
```

## The 14 compiled-only regressions (priority — codegen bugs)

These pass in the interpreter but fail compiled. Three themes:

[...]

Issue #74, opened 2026-04-24
Test262 runner: compile-mode cooperative cancellation


nickna:
## Motivation

Surfaced during the #69 rollout of `test/built-ins/Promise` (commit f869e84). The Test262 runner's compile-mode code path cannot enforce its per-test timeout because emitted IL doesn't check the cancellation token.

Current state — `SharpTS.Test262/Test262Runner.InvokeCompiledMain`:

```csharp
var task = Task.Run(() => mainMethod.Invoke(null, null));
while (!task.IsCompleted) {
    if (task.Wait(100)) break;
    if (DateTime.UtcNow > deadline)
        return new Test262Result(Test262Outcome.Timeout, ...);   // returns, Task keeps running
}
```

The inline comment acknowledges it:

> *"The orphaned Task can't be killed cooperatively (compiled code doesn't honor a cancellation token), so on watchdog fire we accept the memory pressure and move on; the process will GC opportunistically."*

On a small subset this is tolerable. On `test/built-ins/Promise` (677 tests, ~300 of which build unresolvable promise chains) it accumulates hundreds of zombie tasks, testhost creeps past 1 GB, and the full baseline run takes 40+ minutes instead of the expected ~9 minutes. During the rollout this forced holding `test/built-ins/Promise` out of the compiled subset via `compiledExcludeFolders`. Each further folder (`RegExp`, `Object`, `test/language/statements/for`, etc.) is likely to hit the same wall — async-done-deferred tests are skipped up-front, but Timeout-class tests compound.

The interpreter does not have this problem: it polls `CancellationToken.IsCancellationRequested` between statements and at loop boundaries. When the runner calls `cts.Cancel()`, the interpreter throws and unwinds within milliseconds.

## Goal

Compile mode gains an equivalent cooperative-cancellation path so the runner's 5-second per-test timeout actually kills the work. After this lands:

1. Drop `test/built-ins/Promise` from `compiledExcludeFolders`; regenerate compiled baseline; expect a clean ~9-minute run with Timeout bucket populated (not a testhost hang).
2. Future rollout steps (`RegExp`, `Object`, language/statements) add to the compiled subset without the same mitigation.

## Design

Three pieces, all on the compile side. No reflection back to `SharpTS.dll` — standalone-DLL constraint preserved.

1. **Emit cancellation infrastructure into `$Runtime`.** Add a `volatile bool _cancelRequested` static field and a `CheckCancellation()` method to the emitted `$Runtime` type:

   ```csharp
[...]

Issue #856, opened 2026-06-20
Perf epic: close the compiled-output gap to Node (without regressing interop or conformance)


nickna:
## Goal

Make SharpTS **compiled** output **meet or exceed Node.js** on the cross-runtime benchmark suite (`benchmarks/scripts/`), **without** regressing two things we equally care about:

1. **.NET interop** — boxed `object`/`RuntimeValue` interchange, reflective dispatch, and the `$Runtime`/`$Array`/`$TSFunction` host types stay usable from .NET.
2. **TypeScript language conformance** — the Test262 and `microsoft/TypeScript` conformance baselines must not regress. The hot paths below (closures, captures, arrays) are exactly where this repo has had its thorniest *correctness* bugs (write-capture, block-scope shadows, per-iteration cells), so every codegen change here is gated on the full xUnit + conformance suites.

This is a **map of where the time goes**, derived from disassembling the emitted IL of each benchmark workload (`ilspycmd -il`). It is deliberately diagnosis-first: some children have a clear fix, others name where to investigate.

## Where we stand (compiled vs Node, mean ms at largest size)

| Benchmark | Compiled | Node | Slowdown | Dominant cost (from IL) |
|---|---|---|---|---|
| **strings** @10k | 9.85 | 0.084 | **117×** | `s` local typed `object` → dynamic `$Runtime::Add` + O(n²) fresh-string concat |
| **closures** @100k | 101.4 | 1.05 | **97×** | per-call `$TSFunction` build via `GetMethodFromHandle` reflection (slower than our own interpreter!) |
| count-primes @100k | 29.8 | 1.07 | 28× | array as `List<object>`; inline `isinst` ladder + box per element op |
| array-methods @100k | 61.5 | 2.64 | 23× | reflective callback dispatch per element; List↔$Array round-trips |
| objects @100k | 4.2 | 0.30 | 14× | object literal = `Dictionary<string,object>` + string-hash + box per field |
| factorial @10k | 0.013 | 0.0097 | 1.35× | ✅ near parity (tight numeric loop) |
| fibonacci @35 | 34.1 | 73.5 | **0.46× (we win)** | ✅ recursion/call core is already excellent |

[...]

PR #939, opened 2026-06-25
perf(#928): graduate native-int loop counter to default-on + byte-level TypedArray bulk methods

nickna:
Two completable, interop-safe follow-ups from the #928 investigation. The issue's own two levers (native-int element arithmetic; `double[]`/SIMD backing) remain dead-ends for an AOT compiler — the first is unsound (TS `number` is a double, no deopt tier), the second breaks `ArrayBuffer`/`DataView`/`Atomics` aliasing. The dominant cost was always the loop induction variable, which PR #938 addressed behind a flag.

## ① Graduate the native-int loop counter to default-on

PR #938 landed the native-int loop counter gated behind `SHARPTS_INT_LOOP_COUNTER`, **off by default**. It's the biggest lever for compiled typed-array kernels (2 of 3 benchmark kernels meet-or-beat Node with it on). This flips the gate to **default-on**, keeping `SHARPTS_INT_LOOP_COUNTER=0` as a kill-switch.

**Soundness / interop unchanged:** the counter is a non-escaping local and the typed-array `byte[]` backing is untouched. Every value-read still materializes as `conv.r8` (bit-identical to the double counter); only the increment and `a[i±k]` index sites consume the native int.

## ② Byte-level TypedArray bulk methods (interpreter)

The interpreter's `fill`/`set`/`copyWithin`/`reverse`/`slice` boxed every element through the `object?` indexer. They now operate on the `byte[]` backing directly (`Buffer.BlockCopy` / `Span.CopyTo` memmove / exponential-doubling fill / in-place block-swap reverse / a shared `BlockCopyElementsInto` for slice). Interop- and conformance-safe (same backing), and slightly *more* correct — a byte copy preserves NaN bit patterns the old double round-trip could canonicalize.

**Scope: interpreter only.** The compiled emitter (`$TypedArray`) doesn't implement these bulk methods at all (calling them throws `TypeError: object is not a function`), so a dual-mode test would fail in compiled mode. Closing that compiled gap is separate, larger work (filed as a follow-up).

These methods previously had **no dedicated test coverage**; this adds `TypedArrayBulkMethodTests` (18 interpreter tests).

## Validation

- **Full unit suite 14240 / 0** with the counter default-on (incl. `ILVerificationTests`); the suite was also 14222/0 with the counter forced on before ② landed.
- **Compiled differential probe** (f64 stencil, int32 kernel, decrementing loop, counter observed as a string / object key): **byte-identical** output with the counter on vs off.
- **Test262** loop-heavy subset (Array/Math/Number/String/language): counter-on **≡** counter-off, zero delta. (Note: that compiled-Test262 gate is currently red on `main` from a pre-existing stale baseline (#882) and a pre-existing host crash — both reproduce identically with this change reverted, so they are not introduced here.)

## Not done (deliberately)

- Compiled-mode TypedArray bulk-method support — separate follow-up.
- The residual gap vs the idiomatic-C# ceiling traces to the per-iteration `Volatile.Read` cancellation poll at each loop head (#74/#198), a correctness feature, not touched here.

### DIFF
--- a/Compilation/ForLoopAnalyzer.cs
+++ b/Compilation/ForLoopAnalyzer.cs
@@ -18,14 +18,16 @@ namespace SharpTS.Compilation;
 public static class ForLoopAnalyzer
 {
     /// <summary>
-    /// Integer loop-counter prototype gate (#928). When enabled, a provably-integer monotonic
+    /// Integer loop-counter optimization gate (#928). When enabled, a provably-integer monotonic
     /// loop counter is backed by a native <c>Int64</c> slot instead of a <c>double</c> — its
     /// increment stays native int and recognized index sites (<c>a[i]</c>, <c>a[i±k]</c>) consume
-    /// it directly, closing the typed-array kernel gap to Node. Off by default (zero behavior
-    /// change); opt in with <c>SHARPTS_INT_LOOP_COUNTER=1</c> while the prototype is validated.
+    /// it directly, closing the typed-array kernel gap to Node. On by default; it is sound (every
+    /// value-read materializes back to <c>double</c> via <c>conv.r8</c>, bit-identical to the
+    /// double counter, and the byte[] backing is untouched, so .NET interop is preserved). Set
+    /// <c>SHARPTS_INT_LOOP_COUNTER=0</c> to disable it as a kill-switch.
     /// </summary>
     public static readonly bool IntegerCounterEnabled =
-        Environment.GetEnvironmentVariable("SHARPTS_INT_LOOP_COUNTER") == "1";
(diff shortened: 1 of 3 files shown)


nickna:
Filed the compiled-mode follow-up as #940 — TypedArray bulk methods (`fill`/`set`/`copyWithin`/`reverse`/`slice`) are unimplemented in compiled output (they throw `object is not a function`), which is why ② is interpreter-only here. It's linked under the typed-array perf epic #856, and includes the plan to promote `TypedArrayBulkMethodTests` to dual-mode once compiled support lands.

Issue #927, opened 2026-06-25
Unify number[] array representation: complete X3/X4 (delete List<double>, single numeric $Array) without the fresh-append regression


nickna:
## Summary

Follow-up to the escaping `number[]` unboxing work shipped in **#918**. That PR gave compiled `$Array` an unboxed PACKED_DOUBLE elements-kind for **escaping** `number[]` (module-level / function-local / class-field), closing the ~73× write-boxing gap (idx-write ~23× vs boxed ≈2.2× Node; push ~10×, faster than Node). It deliberately stopped at a **two-representation** state:

- numeric `$Array` (`double[]+count`) — for **escaping** arrays
- `List<double>` — for **non-escaping promotable locals** (`ArrayLocalPromotionAnalyzer`)

The end-goal (Nick's) is to **unify on the numeric `$Array` representation and delete the `List<double>` path** (X3/X4 in the design). This issue tracks that.

## Why it isn't done yet (X3/X4 attempted + reverted in #918)

Measured: numeric `$Array` does **not** strictly dominate `List<double>`.
- Steady-state reused array (write+read): numeric ~215ms < List<double> ~257ms (~15% faster ✓)
- **Fresh-per-call append-heavy local** (`function f(){ const a:number[]=[]; for(i) a[i]=i; }`, 1M×20): numeric ~463ms > List<double> ~314ms (**~47% slower ✗**)

Root cause of the append loss: `List<double>` is a tuned path — the array-hoist optimization lifts the `as List<double>` cast out of the loop (no per-write isinst) and BCL `List.Add` is a fast append. Numeric pays per-write dispatch (isinst+castclass+callvirt `SetDouble`) + a branchy `SetDouble` append + the extra `$Array`-wrapper alloc.

## Work required to complete the unify perf-neutrally

1. **Hoist-for-`$Array`**: teach the array-hoist to hoist `as $Array` + call `SetDouble`/`GetLong`, killing per-write isinst. 3 IL sites: hoist setup `ILEmitter.Statements.cs:716`, hoisted set `Properties.cs:1137`, hoisted get `Properties.cs:905`. (Hoist must currently **exclude** Double-kind or it nulls → generic boxed SetIndex = 3200ms.)
2. **`SetDouble`-append tuning** (Unsafe / geometric) to match `List.Add`.
[...]

PR #956, opened 2026-06-26
perf(compile): tune numeric $Array append — dedicated PushDouble + inlining (#927 step 2)

nickna:
Step 2 of #927 — narrows the fresh-append gap between the escaping numeric `$Array` (#918 / #927 step 1) and the non-escaping `List<double>` local path. A standalone win for escaping `number[]`; lands regardless of the (now-declined) unify decision.

## Changes (one file, `RuntimeEmitter.TSArray.cs`)

- **`AggressiveInlining`** on the packed accessors `GetDouble`/`SetDouble`/`PushDouble`. They're non-virtual (`HideBySig`), so a `callvirt` on a `$Array`-typed receiver devirtualizes; the JIT can now fold the field loads + bounds check into the caller's loop — the `$Array` equivalent of `List<double>.Add`'s inlined fast path.
- **Dedicated numeric append in `PushDouble`.** It previously delegated to `SetDouble`, which re-checked `_isNumeric` and ran two dead index branches (`index<_numCount`, `index>_numCount`) on every push. The numeric branch is now straight-line ensure-capacity + store + `_numCount++`/`_length=`, mirroring `List<double>.Add`. Boxed-mode push still delegates to `SetDouble` (rare path).

## Measured (this machine, min-of-5, 1M × 20)

| append form | `List<double>` | numeric baseline | numeric (this PR) |
|---|---|---|---|
| `a.push(i)` | 172 ms | 239 ms (1.34×) | **195 ms (1.13×)** |
| `a[i]=i`    | 213 ms | 262 ms (1.27×) | **249 ms (1.17×)** |

## Correctness

- IL verifies (`--verify`).
- Push / frozen-push-noop / multi-resize growth / push-then-overwrite / aliasing-through-`any` all behave correctly (compiled output spot-checked).
- Full xUnit suite green (confirming in a comment).

## Gate note (#927 step 3)

Numeric still does **not** dominate `List<double>` — the residual ~13–17% is structural (escaping arrays must check frozen/extensible per mutation and live in an `object` slot; provably-non-escaping `List<double>` locals skip both). So deleting `List<double>` (step 3) is not perf-neutral; #927 is closed as won't-unify. This PR is the last of the escaping-array perf wins.

### DIFF
--- a/Compilation/RuntimeEmitter.TSArray.cs
+++ b/Compilation/RuntimeEmitter.TSArray.cs
@@ -295,6 +295,13 @@ public partial class RuntimeEmitter
         var pushDouble = typeBuilder.DefineMethod("PushDouble",
             MethodAttributes.Public | MethodAttributes.HideBySig, _types.Void, [_types.Double]);
         runtime.TSArrayPushDouble = pushDouble;
+        // #927 step 2: these are the per-element hot paths the compiler emits at statically-number[]
+        // sites. They are non-virtual (HideBySig), so a `callvirt` on a $Array-typed receiver
+        // devirtualizes; AggressiveInlining lets the JIT fold the field loads + bounds check into the
+        // caller's loop (the $Array-wrapper equivalent of List<double>.Add's inlined fast path).
+        getDouble.SetImplementationFlags(MethodImplAttributes.AggressiveInlining);
+        setDouble.SetImplementationFlags(MethodImplAttributes.AggressiveInlining);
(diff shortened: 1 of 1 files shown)


nickna:
Full xUnit suite on this branch (step 2 rebased onto current `main`): **14328 / 14329 passed**. The lone failure is `DnsRecordTypeTests.LiveSmoke_ResolveNs_Promise` — a live-network smoke test (returns empty without DNS in the sandbox), pre-existing and unrelated to this change. IL verifies; push correctness re-confirmed vs Node. Ready for review.

Issue #1104, opened 2026-06-29
events/util/assert: dual-mode test coverage + type reconciliation


nickna:
Part of #1097. **types & tests.**

## Problem

The new exports across events/util/assert need test coverage in both modes, and any `BuiltInModuleTypes` entries not covered by facade type-inference need reconciliation.

## Target

- Dual-mode (`interp == compiled`) tests for every new export from the other children of #1097 (parity is structural for facades, but lock it down + guard regressions).
- Verify the new exports type-check (embedded-facade types are inferred from the `.ts` source; add/adjust `GetEventsModuleTypes`/`GetUtilModuleTypes`/`GetAssertModuleTypes` only where inference doesn't cover them).

## Acceptance

- Idiomatic Node usage of the new `events`/`util`/`assert` surface type-checks and runs in both modes.
- `dotnet test`, Test262, TSConformance green (no baseline drift).

Labels: `enhancement`. Ongoing across the epic; final polish last.

Issue #1103, opened 2026-06-29
assert: assert.strict namespace + CallTracker + partialDeepStrictEqual


nickna:
Part of #1097. **assert.**

## Problem

`assert.strict` (the strict-mode namespace), `CallTracker`, and `partialDeepStrictEqual` are absent.

## Target (pure-TS facade)

- **`assert.strict`** — a namespace where the loose forms (`equal`/`deepEqual`/`notEqual`/`notDeepEqual`) alias the strict ones; `assert.strict` is itself callable (=== `assert.ok`) and carries the full member set. (`import { strict as assert } from 'assert'` is a very common pattern.)
- **`CallTracker`** — `calls(fn[, exact])`, `getCalls(fn)`, `report()`, `verify()`, `reset([fn])` (deprecated in Node but documented).
- **`partialDeepStrictEqual(actual, expected[, message])`** (Node 22+) — actual must deep-strict-contain expected.

## Acceptance

- Dual-mode tests: `assert.strict.equal(1,1)` passes / `assert.strict.equal(1,'1')` throws; a `CallTracker.verify()` failure throws; `partialDeepStrictEqual({a:1,b:2},{a:1})` passes — interp == compiled.
- Standalone preserved (pure TS).
- `dotnet test`, Test262, TSConformance green.

Labels: `enhancement`. Depends on the callable-`assert` child.

Issue #1102, opened 2026-06-29
assert: callable assert() + rejects/doesNotReject + match/doesNotMatch + ifError + loose deepEqual


nickna:
Part of #1097. **assert.**

## Problem (code-verified)

`stdlib/node/assert.ts` is missing common members, and — most importantly — **the module isn't callable**: the default export is a plain object literal (`:288`), so `assert(value)` (the most common usage) doesn't work; only `assert.ok(value)` does.

## Target (pure-TS facade)

- **Make `assert` callable**: `assert(value[, message])` === `assert.ok` (a function object carrying all the `assert.*` members).
- **async**: `rejects(asyncFnOrPromise[, error][, message])`, `doesNotReject(...)`.
- **`match(string, regexp[, message])`**, **`doesNotMatch(...)`**.
- **`ifError(value)`** (throws if value is not null/undefined).
- loose **`deepEqual`/`notDeepEqual`** (the `deepEquals` helper at `:88` already takes a `strict` flag — expose the loose exports).

## Acceptance

- Dual-mode tests: `assert(true)` / `assert(false)` throws; `await rejects(async()=>{throw e})`; `match('abc',/b/)`; `ifError(null)` passes / `ifError(err)` throws; `deepEqual({a:1},{a:'1'})` (loose) — interp == compiled.
- Standalone preserved (pure TS).
- `dotnet test`, Test262, TSConformance green.

Labels: `bug`, `enhancement` (the non-callable `assert` is a real defect).

Issue #1101, opened 2026-06-29
util: util.types completeness + MIMEType + abort helpers + parseEnv/getCallSites + legacy aliases


nickna:
Part of #1097. **util.**

## Problem

`util.types` (`util.ts:532`) is partial, and several helpers are missing: `MIMEType`/`MIMEParams`, abort helpers, `parseEnv`/`getCallSites`, and the deprecated legacy aliases.

## Target (pure-TS facade)

- **`util.types.*` completeness**: `isProxy`, `isExternal`, `isAnyArrayBuffer`, `isArrayBufferView`, `isBigIntObject`/`isBooleanObject`/`isNumberObject`/`isStringObject`/`isSymbolObject`, `isNativeError`, `isGeneratorObject`/`isGeneratorFunction`/`isAsyncFunction`, `isModuleNamespaceObject`, `isBoxedPrimitive`, `isKeyObject`/`isCryptoKey`, typed-array predicates — fill the gaps vs the current set.
- **`MIMEType`/`MIMEParams`** classes.
- **`aborted(signal, resource)`** (Promise), `transferableAbortController()`/`transferableAbortSignal(signal)`.
- **`parseEnv(content)`**, **`getCallSites([frames][, options])`** (Node 22+).
- Deprecated aliases for compat: `_extend`, `isArray`/`isBoolean`/`isNull`/`isNullOrUndefined`/`isNumber`/`isString`/`isSymbol`/`isUndefined`/`isObject`/`isFunction`/`isPrimitive`/`isRegExp`/`isDate`/`isError`/`isBuffer`.

## Acceptance

- Dual-mode tests: `util.types.isProxy(new Proxy({},{}))`; `new MIMEType('text/html;charset=utf-8').params`; `await aborted(signal, {})` resolves on abort — interp == compiled.
- Standalone preserved.
- `dotnet test`, Test262, TSConformance green.

Labels: `enhancement`.

PR #1187, opened 2026-07-01
Epic #1097: Node.js events/util/assert facades → 100% compatibility

nickna:
## Epic #1097 — Node.js events/util/assert facades → 100% compatibility

Completes epic #1097 (children #1098–#1104): a completeness sweep of the three most-used pure-logic stdlib facades. Mostly pure-TS additions to `stdlib/node/{events,util,assert}.ts` (interpreter↔compiled parity + standalone automatic), with one exception noted below.

### events
- **#1098** — module-level statics/helpers: `once` (promise), `on` (async iterator, honors `{ signal }`), `getEventListeners`, `setMaxListeners`/`getMaxListeners`, `addAbortListener` (disposable), static `listenerCount`; exposed both as named exports and `EventEmitter` statics.
- **#1099** — `captureRejections` (async-listener rejection → `'error'`), the `errorMonitor` symbol, throw-on-unhandled-`'error'`, plus `EventEmitterAsyncResource` and the `captureRejectionSymbol`/`errorMonitor` symbol exports.
  - **Not pure-TS:** a direct `new EventEmitter()` is backed by the C# `SharpTSEventEmitter` (interpreter) / emitted `$EventEmitter` IL (compiled), not the facade class. So the instance behaviors were implemented in **both** runtime types for true parity. `captureRejections` is opt-in (default off → zero behavior change); throw-on-unhandled is gated to direct instances so internal subclasses (streams/http/net/process/…) stay lenient.

### util
- **#1100** — `styleText`, `formatWithOptions`, `debuglog`/`debug` (NODE_DEBUG-gated), `getSystemErrorMessage`, and `inspect` option completeness (colors/depth/breakLength/maxArrayLength/maxStringLength/showHidden + custom inspection via `Symbol.for('nodejs.util.inspect.custom')`).
- **#1101** — `util.types.*` completeness (isAnyArrayBuffer/isArrayBufferView/typed-array element predicates/isGeneratorFunction/…), `MIMEType`/`MIMEParams`, `aborted`/`transferableAbortController`/`transferableAbortSignal`, `parseEnv`/`getCallSites`, and the deprecated legacy aliases (`_extend`, `isArray`, …).

### assert
- **#1102** — the module is now **callable** (`assert(value)` === `assert.ok`; was a plain object — a real defect), plus `rejects`/`doesNotReject`, `match`/`doesNotMatch`, `ifError`, and loose `deepEqual`/`notDeepEqual`. `looseEquals` now implements JS abstract (`==`) equality explicitly (SharpTS's own `==` doesn't coerce), and pre-existing facade type diagnostics were resolved so `assert` type-checks cleanly via the CLI.
- **#1103** — `assert.strict` (callable namespace; loose forms alias strict), `CallTracker`, `partialDeepStrictEqual`.

### tests & types (#1104)
Dual-mode (`interp == compiled`) tests accompany every child. Facade types flow from the `.ts` source, so no `BuiltInModuleTypes` reconciliation was needed (verified: all new exports type-check via the CLI).

### Known bounds (documented, not chased)
- Legacy global `EventEmitter.captureRejections = true` shares the facade/runtime static split that `defaultMaxListeners` already has — use the per-instance option (works in both modes).
- A compiled facade-subclass `emit(name, ...args)` does not forward args (pre-existing `.apply` limitation; `EventEmitterAsyncResource` tests use no-arg emit).
- `getCallSites` returns `[]` (no guest-visible V8 stack); `isProxy`/boxed-primitive predicates are honestly `false` (transparent / not represented in SharpTS).
- `util.inspect` exact byte layout of exotic objects is approximate (per epic ceiling).

### Validation
- `dotnet test` — **14913/14913** green.
- `SharpTS.TypeScriptConformance` — **31/31** green, no baseline drift.
- `SharpTS.Test262` — SmokeTest 20/0 (baseline) and the interpreted baseline-diff green. The compiled baseline-diff exhibits the documented environmental host-crash flakiness under load; my changes produce byte-identical IL for every Test262 program (none import a node module or use EventEmitter), so there is no compiled-Test262 impact.

Every change is guarded to be a no-op for existing code paths (captureRejections default off; throw-on-unhandled gated to direct EventEmitter instances; all facade additions are new exports).


nickna:
### Compiled Test262 baseline — confirmed pre-existing flake, not a regression

Ran the compiled Test262 baseline on `main` (this branch's base, zero epic changes) as a control:

- **main**: `Failed 1/0` on `test/built-ins/String/prototype/indexOf/searchstring-tostring-bigint.js` (`Fail → RuntimeError`) — a flaky `String.prototype.indexOf`/BigInt entry, entirely unrelated to events/util/assert or EventEmitter.
- **this branch**: the run hit the documented host-crash-under-load flake.

Both are the pre-existing compiled-Test262 instability (memory documents "flaky new-passes that vary run-to-run" + load-induced timeout/crash flips). This epic touches only the events/util/assert facades and the EventEmitter runtime type; **no Test262 program imports a node module or constructs an EventEmitter**, so every Test262 program compiles to byte-identical IL. The interpreted Test262 baseline (same programs, same runtime values) is green.

Issue #1230, opened 2026-07-04
compiled: inner function declaration in a loop body is unreferenceable ("Undefined variable")


nickna:
## Symptom

A `function` declaration inside a **loop body** cannot be referenced as a value in compiled mode — the reference throws at runtime. The interpreter runs the same program correctly.

Found while scoping #1223 (present on baseline before that fix; unrelated to it).

## Repro

```ts
const fns: any[] = [];
function go() {
    for (let i = 0; i < 2; i++) {
        const x = { id: i };
        function h() { return 'x.id=' + x.id; }
        fns.push(h);
    }
}
go();
console.log(fns[0]());
console.log(fns[1]());
```

- Interpreter: prints `x.id=0` / `x.id=1` ✔
- Compiled: `Unhandled exception. System.Exception: ReferenceError: Undefined variable 'h'.` at `$Program.go()` ✗

## Notes

- The failure is at the *reference* site (`fns.push(h)`), suggesting the inner-function lifter registers `h` under a scope the loop-body block doesn't resolve, or declines the lift without a fallback binding.
- Expected semantics (and what a correct fix should preserve): block-scoped function declarations are per-iteration bindings, so each pushed `h` closes over that iteration's `x` (`0`, then `1`).

Issue #1105, opened 2026-06-29
Regex / super / dynamic import silently compile to null in async & generator bodies


nickna:
**Location:** `Compilation/ExpressionEmitterBase.cs:2130` (`EmitRegexLiteral` base = `Ldnull`); `Compilation/GeneratorMoveNextEmitter.Expressions.cs:12-24` (`EmitSuper`/`EmitDynamicImport` → null). Real impl only at `ILEmitter.Expressions.cs:852`.

**Problem (verified firsthand):** the four state-machine emitters derive from `StatementEmitterBase`/`StateMachineExitRoutingEmitter`, never `ILEmitter`, so they inherit base expression stubs that push `null`. A regex literal — and `super.x` as a value, and `import()` — inside an `async function`/`function*` body silently evaluates to `null` at runtime: wrong result, no error. The "base default pushes null, override only in ILEmitter" pattern guarantees every future expression kind silently miscompiles here.

**Action:** make the base virtuals `throw` a clear `CompileException` (fail loud) and port `EmitRegexLiteral` into the shared base (it needs no sync-only state) so regex works in all bodies. File feature-gap follow-ups for `super`/dynamic-import in state machines.

Part of #1088

Issue #307, opened 2026-06-11
compiled: inner function declaration inside an arrow captures enclosing locals by snapshot (reads null/stale) — true root cause of #302 lodash failure


nickna:
## Summary

In **compiled** mode, an inner **function declaration** nested inside an **arrow function / function expression** captures variables from the enclosing scope by **value-snapshot taken at hoist time**, instead of by **live reference**. Because function declarations are hoisted to the top of the enclosing body, the snapshot is taken *before* the captured variable is assigned, so the closure reads back `null` (or a stale value). The **interpreter is correct** (live capture).

This is the real root cause of #302 (compiled lodash init failing in `getNative` with `"object is not a function"`). The original #302 hypothesis — a value-form dispatch gap in `getNative` — was **incorrect**: `getNative` and friends are function declarations inside lodash's `runInContext` (a function expression, compiled as an arrow), and the helpers they reference (`getValue`, `Object`, `symToStringTag`, …) read back `null`. `typeof null === "object"`, so the later `Object(value)` / `pattern.test(...)` calls throw `"object is not a function"`.

## Minimal repro

```js
const ric = () => {
  const Obj = 42;
  function bgt() { return Obj; }   // function DECLARATION inside an arrow, captures Obj
  return bgt;
};
console.log(ric()());
```

```
$ dotnet run -- repro.ts                         # interpreter
42
$ dotnet run -- --compile repro.ts -o r.dll && dotnet r.dll
null                                             # WRONG: Obj snapshot taken before `const Obj = 42` ran
```

## Why

`function` declarations are hoisted: `EmitInnerFunctionHoisting` runs at the **top** of the arrow body (`ILCompiler.ArrowFunctions.cs` ~1227), *before* the body statements that assign `Obj`. Its Pass-2 (`ILCompiler.InnerFunctions.cs` ~501) **copies the captured value** into the inner function's own display-class field. At that point `Obj` has not been assigned, so the field is the zero/`null` default.

The two cases that *do* work avoid the snapshot:
- **All arrows** (`const bgt = () => Obj;`): the inner arrow reads `Obj` **live** through the enclosing arrow's scope DC (`$arrowDC` reference), not a copy.
- **All function declarations** (`function ric() { … function bgt() { return Obj; } }`): the inner function reads `Obj` **live** through the enclosing function's display class (`$functionDC` reference).

[...]

PR #1239, opened 2026-07-04
fix(compile): #1237 materialize inner function declarations inside class methods

### LINKED ISSUE #1237: compiled: inner function declarations inside a class method are unreferenceable
## Symptom

A `function` declaration inside a **class method** body cannot be referenced (or even called) in compiled mode — the reference throws `ReferenceError: Undefined variable` at runtime. The interpreter runs the same program correctly. This affects both non-capturing and capturing inner functions, and both top-level-in-method and block-nested declarations.

Found while fixing #1230 (inner function in a loop body inside a plain function). That fix wires block-scoped inner-function materialization into the function / arrow / inner-function body emitters, but the **class-method emission path is separate** and never hoists inner function declarations at all.

## Repro

```ts
class C {
    m() {
        function top() { return "top"; }
        return top();
    }
}
console.log(new C().m());
```

- Interpreter: prints `top` ✔
- Compiled: `Unhandled exception. System.Exception: ReferenceError: Undefined variable 'top'.` at `C.m()` ✗

Capturing and loop-body variants fail the same way:

```ts
class C {
    m() {
        const k = "K";
        function cap() { return "k=" + k; }   // top-level capturing
        return cap();
    }
}
```

## Root cause

`ILCompiler.Classes.Methods.cs` emits a method body via a bare `foreach (var stmt in method.Body) emitter.EmitStatement(stmt);` and never calls `EmitInnerFunctionHoisting` (the pre-pass that the plain-function path at `ILCompiler.Functions.cs`, arrow bodies, and inner-function bodies all run). So an inner `function` in a method is collected (its method + display class are emitted) but never materialized into a binding, and every reference falls through to `ThrowUndefinedVariable`.

## Fix direction

Give class-method bodies the same treatment the function/arrow paths got in #1230: run the top-level `EmitInnerFunctionHoisting` pre-pass over `method.Body` (which also wires `ctx.EmitBlockScopedInnerFunction` so block-nested declarations materialize in place with correct per-iteration captures). Needs verification that the method ctx carries the fields the hoisting/population reads (`FunctionDisplayClass*`, `CapturedFunctionLocals`, entry-point DC, etc.), and that this composes with async/generator method state-machine emission (which uses separate emitters).

nickna:
## Summary

Fixes #1237. A `function` declaration inside a **class method** body was unreferenceable in compiled mode — every reference threw `ReferenceError: Undefined variable` at runtime, while the interpreter ran the same program correctly. This affected non-capturing, capturing, top-level-in-method, and block-nested inner functions, on instance, static, and private methods.

## Root cause

The three sync class-method emitters — `EmitMethod` and `EmitPrivateMethodBody` (`ILCompiler.Classes.Methods.cs`) and `EmitStaticMethodBody` (`ILCompiler.Classes.Static.cs`) — emit the body via a bare `foreach (var stmt in method.Body) EmitStatement(stmt)` and never wired the inner-function materializer that the plain-function (`ILCompiler.Functions.cs`) and arrow paths run. So an inner `function` in a method was **collected** (its method and display class were emitted) but never **materialized into a binding**, and every reference fell through to `ThrowUndefinedVariable`.

## Fix

Adds `WireInPlaceInnerFunctions(ctx)` (`ILCompiler.InnerFunctions.cs`) and calls it before the body loop in all three method paths. It wires the #1230 in-place materializer (`EmitBlockScopedInnerFunctionDeclaration`) onto the method context but deliberately leaves `HoistedInnerFunctions` **empty**, so every inner function declaration — top-level-of-body or block-nested — is created at its textual position by the statement emitter's `Stmt.Function` arm.

Why in-place rather than the function/arrow path's top-of-body two-pass hoist: unlike a plain function or arrow, a method has **no function-level display class**, so hoisting a capturing inner function to the top of the body would snapshot its captured method-locals *before* the body assigns them (null capture, e.g. `k=` where `k=K` is expected). In-place materialization snapshots after the preceding statements run, so captures are correct — matching how arrows inside methods already capture. In-place is also *sufficient*: the type checker rejects forward references to a method's inner functions (mutual recursion / call-before-declaration both error), so no valid program needs the two-pass pre-materialization here.

3 files, +38 lines.

## Tests

New `SharpTS.Tests/SharedTests/InnerFunctionInMethodTests.cs` — 11 theories × both execution modes (22 cases): non-capturing, method-local const, parameter (value-type double boxing), numeric local (boxing), self-recursion, two inner functions sharing a capture, module top-level-var capture (entry-point DC), loop-body per-iteration, if-block block-const, static method, private method.

## Verification

- **xUnit:** 15404 / 0 (incl. the 22 new tests)
- **Test262 (interpreter + compiled):** 22 / 0 / 4 skip — baseline-green
- **TypeScript conformance:** 31 / 0
- **IL `--verify`:** passes on all repros

## Out of scope (pre-existing, verified not regressions)

- An inner function *returning an arrow* that captures an outer local reads it as `0` in compiled mode. The identical structure **crashes** (`NullReferenceException`) at plain-function scope too — a pre-existing #307/#313 arrow-scope-DC-through-inner-function-boundary threading bug, not method-specific. This fix makes the method case fail softer (wrong value, no crash).
- Inner functions in **async/generator** methods use separate state-machine emitters (untouched) and remain broken (#1105-class silent-null).
- The type checker hoists forward references in *function* bodies but not *method* bodies — a separate type-checker inconsistency, not addressed here.

### DIFF
--- a/Compilation/ILCompiler.Classes.Methods.cs
+++ b/Compilation/ILCompiler.Classes.Methods.cs
@@ -811,6 +811,10 @@ public partial class ILCompiler
         // Emit method body
         if (method.Body != null)
         {
+            // #1237: materialize inner function declarations in place, matching the public-method
+            // path so an inner `function` declared inside a private method becomes a binding.
+            WireInPlaceInnerFunctions(ctx);
+
             foreach (var stmt in method.Body)
             {
                 emitter.EmitStatement(stmt);
@@ -1185,6 +1189,15 @@ public partial class ILCompiler
         // Abstract methods have no body to emit
         if (method.Body != null)
         {
+            // #1237: an inner `function` declared inside a method is collected (its method/display
+            // class are emitted) but was never materialized into a binding here, so every reference
+            // fell through to ThrowUndefinedVariable. Wire the in-place materializer so each inner
+            // function declaration is created at its textual position by the statement emitter's
+            // Stmt.Function arm. Methods have no function-level display class, so a top-of-body hoist
+            // (EmitInnerFunctionHoisting) would snapshot captured method-locals before they are
+            // assigned; in-place materialization captures them correctly. See WireInPlaceInnerFunctions.
+            WireInPlaceInnerFunctions(ctx);
+
             foreach (var stmt in method.Body)
             {
                 emitter.EmitStatement(stmt);
--- a/Compilation/ILCompiler.Classes.Static.cs
+++ b/Compilation/ILCompiler.Classes.Static.cs
@@ -505,6 +505,10 @@ public partial class ILCompiler
         // Abstract methods have no body to emit
         if (method.Body != null)
         {
+            // #1237: materialize inner function declarations in place, matching the instance-method
+            // path so an inner `function` declared inside a static method becomes a binding.
+            WireInPlaceInnerFunctions(ctx);
+
             foreach (var stmt in method.Body)
             {
                 emitter.EmitStatement(stmt);
--- a/Compilation/ILCompiler.InnerFunctions.cs
+++ b/Compilation/ILCompiler.InnerFunctions.cs
@@ -718,6 +718,27 @@ public partial class ILCompiler
     }
 
     /// <summary>
+    /// Wires the in-place inner-function materializer (<see cref="EmitBlockScopedInnerFunctionDeclaration"/>)
+    /// onto <paramref name="ctx"/> WITHOUT running the top-of-body two-pass hoist that
+    /// <see cref="EmitInnerFunctionHoisting"/> performs. Class-method bodies (#1237) use this: unlike a
+    /// plain function or arrow, a method has no function-level display class, so hoisting a capturing
+    /// inner function to the top of the body would snapshot its captured method-locals BEFORE the body
+    /// assigns them — leaving the closure's capture fields null. Leaving
+    /// <see cref="CompilationContext.HoistedInnerFunctions"/> empty means every inner function
+    /// declaration — top-level-of-body or block-nested — is materialized in place at its textual
+    /// position by the statement emitter's <c>Stmt.Function</c> arm, so its capture snapshot sees the
+    /// values already assigned. This matches how arrows inside methods capture (snapshot at creation),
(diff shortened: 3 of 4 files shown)


nickna:
### Follow-up: the CI failure was a real data race, now fixed (commit 3110b3c7)

The `build (ubuntu-latest)` failure on the first push was **not** a flaky test and **not** caused by the inner-function change — it was a genuine pre-existing concurrency bug that this commit fixes.

**Failure:** `ClusterModuleTests.RoundRobin_DistributesConnectionsAcrossWorkers(mode: Compiled)` — a 60s `TimeoutException` (a hang, not an assertion failure), Ubuntu-only, intermittent.

**Root cause:** unsynchronized check-then-act in `SharedListenerRegistry.cs`:

```csharp
if (!_started) { _started = true; _listener.Start(); StartAccepting(); }
```

Under `SCHED_RR`, both cluster workers call `server.listen(samePort)` from their own interpreter threads, and `ConcurrentDictionary.GetOrAdd` hands both the **same** `SharedTcpListener`. When the two `AddWorker` calls overlap, both pass `!_started`, both call `_listener.Start()`, and the second `Bind` on the socket throws. That exception propagates out of `ListenAsClusterWorker` before it fires the `listening` callback / `process.send('ready')`, so the primary's `readyCount` never reaches 2, the 4 connections are never made, and the module hangs to the 60s timeout. Multi-core + CI load widens the race window, which is why it was intermittent and Compiled-only.

**Reproduced locally** on the pre-fix build: `Fork_WorkersShareNetPort(mode: Compiled)` hung 61s on the first loop iteration, next iteration passed in 7s.

**Fix:** wrap the start-once transition in a lock (double-checked; publish `_started` only after `Start()` succeeds) in both `SharedTcpListener` and `SharedHttpListener`, and guard `Stop()` with the same lock.

**Verification:** 40/40 iterations of the three port-sharing tests pass post-fix (no hang), full `ClusterModule`/`Net`/`Http` suite 130/0.

This is logically independent of the #1237 inner-function fix (first commit); it's bundled here because it's what was blocking this PR's CI. Happy to split it into its own PR if preferred.

Issue #1242, opened 2026-07-04
Extract per-receiver branches from EmitGetProperty (follow-up to #1141)


nickna:
Follow-up to #1141 / epic #1094.

PR #1241 split the 5,222-line `RuntimeEmitter.Objects.Properties.cs` at the **file** level (Set*/Delete* emitters moved to sibling partials, leaving 3,090 lines), but deliberately did **not** extract the per-receiver branches *inside* the 2,028-line `EmitGetProperty` method into `Emit<Receiver>GetBranch(il, …)` helpers (the original #1141 ask).

**Why deferred:** `EmitGetProperty` emits IL through method-scoped labels and locals that are defined once at the top and shared/marked across every receiver branch (globalThis, `__proto__`, proxy, namespace, `$Object`, dict, list, string, class-instance, error-type, …). Slicing a branch into a helper means threading those shared labels/locals through the call and risks silently emitting different IL — a class of bug the test suites may not surface for every receiver×property combination.

**Action:** extract branches incrementally, one receiver type at a time, each verified with `--verify` (IL validation) plus the full `dotnet test` + Test262 (compiled mode) gate. Mirror the already-factored `EmitProxyGetPropertyCheck`/`EmitProxyDeleteCheck` shape (pass `il` + arg-loader delegates + the fall-through label). `EmitSetProperty` (now in `RuntimeEmitter.Objects.SetProperty.cs`) has the same internal structure and is a candidate too.

Part of #1094.

Issue #393, opened 2026-06-13
Compiled timers/promises helpers fail IL verification ($M_promises_setTimeout/setImmediate StackUnexpected)


nickna:
### Observed while working on #381.

Any compiled program importing `timers/promises` fails `--verify`:

```ts
import { setTimeout } from 'timers/promises';
await setTimeout(300, 'tick');
```

```
$ dotnet run -- --compile prog.ts -o prog.dll --verify
IL verification found 2 error(s):
  [IL Error] $Program.$M_promises_setTimeout: StackUnexpected - Unexpected type on the stack.
  [IL Error] $Program.$M_promises_setImmediate: StackUnexpected - Unexpected type on the stack.
```

### Notes
- Pre-existing on `main` (confirmed by stashing all local changes and re-running `--verify`); not introduced by #381.
- The program still **runs** correctly (`tick` is printed) — the bad IL is in the emitted `timers/promises` helper methods `$M_promises_setTimeout` and `$M_promises_setImmediate`, which the JIT tolerates but ILVerify rejects.
- Both helpers are emitted regardless of which `timers/promises` member is imported (so importing only `setTimeout` still emits a malformed `$M_promises_setImmediate`).
- Likely a stack-type mismatch (e.g. a value left boxed/unboxed or a wrong cast) in the emitter for these two helpers. Other async surfaces (plain `async function`, single- and multi-module) verify clean.

### Repro
`dotnet run -- --compile <any program importing timers/promises> --verify`

PR #1248, opened 2026-07-04
fix(compile): make the fs facade emit IL-verifiable methods (#1246)

### LINKED ISSUE #1246: Compiled fs facade emits IL-unverifiable methods (68 --verify errors: $Promise vs Task<object>, object vs string) although it runs
## Summary

Compiling any program that imports `fs` produces **68 IL verification errors** (`--verify`) in the emitted `fs` facade module. The program **runs correctly** (the JIT tolerates the mismatches), but the emitted IL is not strictly verifiable. This matters for standalone `--verify` gating and any stricter/AOT runtime.

Found incidentally while running `--verify` during #1242 (a pure `EmitGetProperty` refactor). Reproduced on `main` before that PR — **not** a regression from #1242, and none of the errors are in `GetProperty`/`$Stats`.

## Minimal repro

```ts
import * as fs from "fs";
console.log(fs.statSync(".").isDirectory());
```

```
$ dotnet run -- --compile repro.ts -o repro.dll --verify
Compiled to repro.dll
IL verification found 68 error(s):
  [IL Error] $Program.$M_fs_cpSync: StackUnexpected {Offset=854, Found=ref 'object', Expected=ref 'string'} ...
  ...
$ dotnet repro.dll
true          # runs fine despite the verify errors
```

## Two error classes

**1. `Found ref 'object', Expected ref 'string'`** — synchronous / helper methods and their closure display-classes:
`$M_fs_cpSync`, `$M_fs_mkdirSync`, `$M_fs___makeDirent`, `$M_fs___globWalk`, `$M_fs_globSync`, plus many `<>c__DisplayClassNN.Invoke`. A value typed `object` is left on the stack where a `string` is expected (a missing cast/`castclass` before a string-typed use).

**2. `Found ref '$Promise', Expected ref 'Task`1<object>'`** — the async fs methods:
`$M_fs_readFile`, `writeFile`, `appendFile`, `unlink`, `access`, `mkdir`, `rmdir`, `readdir`, `rename`, `copyFile`, `chmod`, `chown`, `truncate`, `utimes`, `readlink`, `realpath`, `symlink`, `link`, `mkdtemp`, `open`, `close`, `fstat`, `ftruncate`, `fsync`, `fdatasync`, `rm`, `cp`, `glob`, `statfs`, `lchmod`, `lchown`, `lutimes`, `fchmod`, `fchown`, `futimes`, … The facade returns a `$Promise` where the method's declared return / stack slot expects `System.Threading.Tasks.Task<object>` (a missing bridge/cast from `$Promise` to its underlying `Task<object>`).

## Likely cause

The `fs` facade lowering emits values at their erased `object`/`$Promise` type without inserting the `castclass` the surrounding IL expects (`string` for the sync string-path args; `Task<object>` for the async return). This is the known fs/facade compiled-divergence theme; ILVerify is strict where the JIT is lenient.

## Impact

- `--compile … --verify` cannot be used as a clean gate on any fs-using program.
- Runtime behavior is currently unaffected (JIT-tolerated), but a stricter runtime / AOT path could reject these.

## Repro tally

64 of the 68 are the two classes above (the rest are the same two shapes in the display-class `Invoke` bodies). Full list reproducible via the one-liner above.

nickna:
Closes #1246.

## Problem

Compiling any program that imports `fs` produced **68 IL verification errors** in the emitted facade module. The program ran fine (the JIT tolerates the mismatches), but `--compile … --verify` couldn't be a clean gate on any fs-using program.

```
$ dotnet run -- --compile repro.ts -o repro.dll --verify
IL verification found 68 error(s):
  [IL Error] $Program.$M_fs_cpSync: StackUnexpected {Found=ref 'object', Expected=ref 'string'} ...
  [IL Error] $Program.$M_fs_readFile: StackUnexpected {Found=ref '$Promise', Expected=ref 'Task`1<object>'} ...
```

Both shapes are the same underlying defect — a value stored into a narrower slot without the bridge the verifier requires — surfacing in two forms.

## Root causes & fixes

**1. `object` → `string` (sync helpers + display-class `Invoke`s).**
A `$M_fs` helper stages an `any`-typed argument (e.g. `__joinPath(src, names[i])` in `cpSync`, where `names[i]` is an `object`) into a `string` parameter temp. `EmitConversionForParameter` handled only value-type targets (unbox) and fell through for reference targets, leaving `object` on the stack where `string` was expected — no `castclass`. Fixed by emitting `castclass` on string parameter coercion (in both the `ILEmitter` override and the shared base). Restricted to `string`: it's the one non-object reference slot whose runtime value is a genuine CLR instance of the slot. Delegates (`$TSFunction`), dynamic collections (`$Array`/`$Map`/`$Set`), and Task (`$Promise`) carry a boxed-differently runtime value, so a `castclass` would throw — they keep their prior (JIT-tolerated) object store.

**2. `$Promise` → `Task<object>` (async callback/promise helpers).**
The promise primitives return a wrapped `$Promise`, which the callback helpers (`__cbData`/`__cbVoid`) took in a `Promise<any>` → `Task<object>` parameter slot. `ParameterTypeResolver.CoerceParamSlotType` already widens other unsound parameter slots to `object` (Date/RegExp/array/Map/Set, unions admitting `undefined`, symbols) — and `ResolveReturnType` widens non-async `Promise<T>` **returns** to object (#393) — but the symmetric widening for `Promise<T>` **parameters** was never added. Fixed by widening `Promise<T>` (→ `Task`/`Task<object>`) parameter slots to `object`, mirroring #393. A `Promise<T>` param always receives the runtime `$Promise` object regardless of whether the enclosing function is async (only the function's *return* builds a real Task), so the widening is unconditional.

With the parameter slots honest (`object`), no runtime bridging is needed and the `$Promise` stores cleanly.

## Testing

- New `ILVerificationTests`: `FsFacadeImport_PassesILVerification` (the issue repro, whole facade), plus a minimal plain-TS repro of each class (`ObjectArgumentToStringParameter_*`, `PromiseValueToPromiseParameter_*`). Each was confirmed to fail pre-fix and pass post-fix.
- **xUnit: 15415/0** · **Test262: 22/0/4skip** (interp + compiled) · **TSConformance: 31/0**
- `--verify` and `--standalone --verify` clean on the repro; interp/compiled output identical across a broad fs exercise (sync + callback + `fs/promises`).

No behavior change for correctly-typed programs — the fix only makes the already-working IL verifiable. Found incidentally while running `--verify` during #1242; not a regression from it.

### DIFF
--- a/Compilation/ExpressionEmitterBase.CallHelpers.cs
+++ b/Compilation/ExpressionEmitterBase.CallHelpers.cs
@@ -2361,9 +2361,25 @@ public abstract partial class ExpressionEmitterBase
         }
 
         // If target is a value type and we have an object, unbox
-        if (targetType.IsValueType && StackType != StackType.Double && StackType != StackType.Boolean)
+        if (targetType.IsValueType)
         {
-            IL.Emit(OpCodes.Unbox_Any, targetType);
+            if (StackType != StackType.Double && StackType != StackType.Boolean)
+                IL.Emit(OpCodes.Unbox_Any, targetType);
+            return;
+        }
+
+        // Target is a non-object reference type. EmitExpression can leave a broader `object` on the
+        // stack — an `any`-typed value, or a GetIndex/GetProperty result — that the callee's typed
+        // string slot would reject at IL verification even though the JIT tolerates it. Emit the
+        // downcast the verifier needs (#1246). Only `string` qualifies: it is the one non-object
+        // reference slot whose runtime value is genuinely a CLR instance of the slot. Other reference
+        // targets carry a boxed-differently runtime value ($TSFunction for a Delegate, $Array for a
+        // List, $Promise for a Task — the last widened to `object` at the parameter slot by
+        // ParameterTypeResolver), so a castclass would throw; they keep their prior object store.
+        if (Types.IsString(targetType) && StackType != StackType.String)
+        {
+            EnsureBoxed();
+            IL.Emit(OpCodes.Castclass, targetType);
         }
     }
 
--- a/Compilation/ILEmitter.Helpers.cs
+++ b/Compilation/ILEmitter.Helpers.cs
@@ -276,9 +276,25 @@ public partial class ILEmitter
         }
 
         // If target is a value type and we have an object, unbox
-        if (targetType.IsValueType && _stackType != StackType.Double && _stackType != StackType.Boolean)
+        if (targetType.IsValueType)
         {
-            IL.Emit(OpCodes.Unbox_Any, targetType);
+            if (_stackType != StackType.Double && _stackType != StackType.Boolean)
+                IL.Emit(OpCodes.Unbox_Any, targetType);
+            return;
+        }
+
+        // Target is a non-object reference type. EmitExpression can leave a broader `object` on the
+        // stack — an `any`-typed value, or a GetIndex/GetProperty result — that the callee's typed
+        // string slot would reject at IL verification even though the JIT tolerates it. Emit the
+        // downcast the verifier needs (#1246). Only `string` qualifies: it is the one non-object
+        // reference slot whose runtime value is genuinely a CLR instance of the slot. Other reference
+        // targets carry a boxed-differently runtime value ($TSFunction for a Delegate, $Array for a
+        // List, $Promise for a Task — the last widened to `object` at the parameter slot by
+        // ParameterTypeResolver), so a castclass would throw; they keep their prior object store.
+        if (_ctx.Types.IsString(targetType) && _stackType != StackType.String)
+        {
+            EmitBoxIfNeeded(expr);
+            IL.Emit(OpCodes.Castclass, targetType);
         }
     }
 
--- a/Compilation/ParameterTypeResolver.cs
+++ b/Compilation/ParameterTypeResolver.cs
@@ -163,8 +163,27 @@ public static class ParameterTypeResolver
             return typeof(object);
         if (UnionAdmitsUndefined(source))
             return typeof(object);
+        if (IsPromiseTaskSlot(mapped))
+            return typeof(object);
         return mapped;
     }
+
+    /// <summary>
+    /// True when <paramref name="mapped"/> is a CLR <c>Task</c>/<c>Task&lt;T&gt;</c> produced by
+    /// strictly mapping a <c>Promise&lt;T&gt;</c> annotation. A parameter so typed receives the runtime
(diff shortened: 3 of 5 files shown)


nickna:
## CI investigation: Ubuntu failure was a pre-existing flaky test (now fixed)

The first CI run failed only on **ubuntu-latest** (windows-latest passed), on a single test: `ProcessGlobalTests.Process_Uptime_IncreasesOverTime(mode: **Interpreted**)` — `Expected "true", Actual "false"`. The **Compiled** variant of the same test passed in the same run.

**This is unrelated to the fix in this PR** (which is IL-compiler-only): the failing test exercises the **interpreter** path, in `Runtime/BuiltIns/ProcessBuiltIns.cs`.

### Root cause
`uptime()` returns `now - (_scriptStartTimestamp ?? _startTimestamp)`. `_scriptStartTimestamp` was a **shared mutable static** that the in-process test runner toggles on every interpreted run — `SetScriptArguments` sets it to a recent timestamp, `ClearScriptArguments` nulls it (in a `finally`). Because xUnit parallelizes test collections across one CLR process, a concurrent script's `SetScriptArguments` can land **between** this test's two `uptime()` reads:

1. `up1` read while the baseline is `null` → falls back to `_startTimestamp` (process start) → **large** elapsed (~80s into the CI run).
2. A concurrent `RunInterpretedWithArgs` calls `SetScriptArguments` → baseline becomes a **recent** timestamp.
3. `up2` read against the recent baseline → **~0** elapsed.
4. `0 >= 80` → `false`. Fail.

`up2 >= up1` uses `>=`, so a mere failure-to-advance wouldn't fail it — uptime genuinely went **backwards**. The earlier "use `Stopwatch`" change (see the code comment naming this exact test) fixed clock monotonicity but not this **baseline-switching** race. It's timing- and interleaving-dependent, which is why Windows passed, Compiled passed, and it's green locally.

### Fix (commit `8114f87`)
Make `_scriptStartTimestamp` `[ThreadStatic]` — the same precedent as the `ThreadArgv`/`ThreadEnv` fields directly above it — so each concurrently-running script gets a stable per-thread baseline and can't stomp another's. `Interpret()` sets the per-thread baseline before any user code runs, so both reads in a script see the same baseline → monotonic. Not a production bug: a real single-script process sets the baseline once on its own thread with no concurrent script to clear it.

Full suite green under the CI filter (`Category!=LiveNetwork&Category!=LoadSensitive`): **15408/0**.

Issue #1001, opened 2026-06-28
worker_threads: emit the 'messageerror' event


nickna:
Part of epic #996.

## Problem (verified)

Node fires `'messageerror'` on the receiving `Worker`/`MessagePort` when an incoming message fails to deserialize. SharpTS clones **eagerly on the sender** and throws `DataCloneError` synchronously to the poster (`Runtime/Types/SharpTSWorker.cs:466`, `Runtime/Types/SharpTSMessagePort.cs:132`), so there is no receiver-side failure point — `'messageerror'` never fires (only `BroadcastChannel` emits it today).

## Fix

Defer/emulate the receiver-side clone failure so `'messageerror'` is emitted on the receiving port (instead of, or in addition to, throwing on the sender), matching Node's delivery model. Decide whether to keep the synchronous sender-side throw for the strictly-uncloneable cases or fully move to receiver-side semantics.

## Tests

Dual-mode: sending a value that fails to materialize on the receiver surfaces as `'messageerror'` on the receiving Worker/MessagePort.

nickna:
Delivered in #1066, merged to `main` (merge commit c7671765). Part of epic #996. (The PR's "Closes" list only auto-closed #997 due to GitHub's single-keyword-per-reference behavior; closing manually.)

Issue #1000, opened 2026-06-28
worker_threads: real environment-data store + receiveMessageOnPort returns undefined


nickna:
Part of epic #996.

## Problem (verified)

- `getEnvironmentData` / `setEnvironmentData` / `SHARE_ENV` are backed by `process.env` (`Environment.Get/SetEnvironmentVariable`, `Runtime/BuiltIns/Modules/Interpreter/WorkerThreadsModuleInterpreter.cs:78-104`). Node keeps a **separate internal environment-data Map keyed by arbitrary values**, snapshotted into each worker at spawn — independent of `process.env`.
- `receiveMessageOnPort` returns CLR `null` when the port queue is empty (`Runtime/Types/SharpTSMessagePort.cs:311`, `ReceiveMessageSync`); Node returns `undefined` (and `{ message }` when present). (The method is already correctly non-blocking.)

## Fix

- Replace the `process.env` backing with a real per-process environment-data Map keyed by arbitrary values; snapshot it into the worker at construction.
- Return `undefined` (not `null`) from `receiveMessageOnPort` on an empty/closed port.

## Tests

Dual-mode: env data set in the parent is visible in the worker via `getEnvironmentData`, and does **not** leak into `process.env`; `receiveMessageOnPort` on an empty port is `undefined`, on a non-empty port is `{ message }`.

nickna:
Delivered in #1066, merged to `main` (merge commit c7671765). Part of epic #996. (The PR's "Closes" list only auto-closed #997 due to GitHub's single-keyword-per-reference behavior; closing manually.)

Issue #996, opened 2026-06-28
Epic: bring `worker_threads` to 100% Node parity


nickna:
## Goal

Bring SharpTS's Node.js `worker_threads` module to **100% Node parity** (of what the .NET runtime can faithfully express), with the interpreter and compiled paths behaving identically — so we can declare `worker_threads` done.

The module is already substantially built: one real .NET thread per `Worker`, structured-clone message passing, real `SharedArrayBuffer` + `Atomics` (genuine multicore), `ref`/`unref` event-loop liveness, `MessageChannel`/`MessagePort`/`BroadcastChannel`, cross-runtime port transfer. **PR #981** separately closed the *interpreter concurrency-isolation* soundness hole (per-realm `globalThis`/prototype `_extras`/`Symbol.for`). This epic tracks the remaining **API + teardown** surface.

## Audit (2026-06-28, code-verified)

Done in worktree `wrk/worker-threads-gap-audit` off clean `main`, against `Runtime/Types/SharpTSWorker.cs`, `SharpTSMessagePort.cs`, `StructuredClone.cs`, `SharpTSAtomics.cs`, `TypeSystem/BuiltInModuleTypes.cs`, and `SharpTS.Tests/SharedTests/WorkerThreadsTests.cs`.

**Already handled (not gaps):**
- Compiled programs that use `Worker` already record `RequireSharpTSRuntime("Worker")` (`ExpressionEmitterBase.Constructors.cs:468`) → the CLI co-locates SharpTS.dll; `--standalone` suppresses and Worker throws a clear error.
- The worker child always runs in a fresh interpreter (`SharpTSWorker.cs:297`), so #981's per-realm isolation already covers worker bodies → compiled concurrent-worker isolation is sound by construction. (The compiled "gap" is fidelity — no pure-compiled worker body — not soundness, and is out of scope here.)

**Helpful:** the `worker_threads` type surface is entirely `any` (`BuiltInModuleTypes.cs:1133`), so new events/methods need **zero** type-checker changes.

## Children (priority order)

- [ ] #997 — **`terminate()` teardown (bug).** Wake a worker parked in `Atomics.wait` (the `_waiters` registry has no per-worker cancel hook) + emit `exit` code `1` on error/terminate. The one genuine correctness bug.
- [ ] #998 — **Emit the `'online'` event.** Never fired today; trivial to enqueue at worker start.
[...]

PR #1253, opened 2026-07-05
feat(worker_threads): compiled main-thread receiveMessageOnPort drain + MessageChannel messageerror (#1077)

### LINKED ISSUE #1077: worker_threads: compiled main-thread MessageChannel messageerror + synchronous receiveMessageOnPort
Two related compiled-mode gaps documented during epic #996 (#1001 and #1000). The **worker-driven** paths already work dual-mode; these are the compiled **main-thread `MessageChannel`/`$MessagePort`** cases.

## 1. Compiled main-thread `receiveMessageOnPort` is a stub (from #1000)

`$Runtime.WorkerThreadsReceiveMessageOnPort` (`Compilation/RuntimeEmitter.Worker.cs`, `EmitWorkerThreadsModuleHelpers`) returns `undefined` unconditionally — it cannot synchronously drain a `$MessagePort` queue because the `$MessagePort` type (and its `_pending`/`_closed` fields) is emitted **after** the worker-threads helpers: `EmitMessageChannelTypes` runs after `EmitRuntimeClass` (`Compilation/RuntimeEmitter.cs:~425` vs the `EmitWorkerHelpers` call inside `EmitRuntimeClass`). So the helper can't reference `$MessagePort` at emit time.

**Fix options:** emit the `receiveMessageOnPort` helper *after* `EmitMessageChannelTypes` (define the method on the runtime type later, set `runtime.WorkerThreadsReceiveMessageOnPort` before the user-code pass), or route through a deferred helper that reflects on the `$MessagePort` `_pending` queue. Then dequeue and return `{ message }` (a `Dictionary<string,object?>` reads correctly via compiled `.message`), else `undefined`.

## 2. Compiled `$MessagePort.postMessage` of an uncloneable value doesn't fire `messageerror` (from #1001)

The emitted `cloneCore` (`Compilation/RuntimeEmitter.Worker.cs:~1672`, `fallbackReturn`) returns uncloneable values (e.g. a function/closure) **by reference** instead of throwing `DataCloneError`, so `$MessagePort.postMessage(uncloneable)` has no clone-failure point and never delivers `messageerror`. (The C# `SharpTSWorker` paths *do* throw, so worker→parent / parent→worker `messageerror` is dual-mode; only the `MessageChannel`/`$MessagePort` path is interpreter-only.)

**Fix options:** make the emitted clone detect strictly-uncloneable values and throw `DataCloneError` (careful — `cloneCore` is shared by compiled `structuredClone`, `$MessagePort`, and `$BroadcastChannel`, so changing it affects those), or detect uncloneable in `$MessagePort.PostMessage` before enqueue and deliver a messageerror marker at `Drain`.

## Tests

Dual-mode: compiled main-thread `receiveMessageOnPort(port)` returns `{ message }` when a message is queued; compiled `port.postMessage(()=>{})` fires `messageerror` on the receiving `MessageChannel` port.

nickna:
Closes #1077.

Fixes the two compiled-mode gaps documented during epic #996 for the **main-thread `MessageChannel`/`$MessagePort`** path. The worker-driven paths were already dual-mode; this brings the interpreter-only `MessageChannel` cases to parity.

## 1. Synchronous `receiveMessageOnPort` (from #1000)

`$Runtime.WorkerThreadsReceiveMessageOnPort` was an unconditional `undefined` stub: `$MessagePort` is emitted by `EmitMessageChannelTypes` **after** `EmitRuntimeClass`, so the helper couldn't reference the port's `_pending` queue at emit time.

**Fix:** the method is still *defined* during `EmitWorkerThreadsModuleHelpers` (so call sites bind the token), but its **body** is now emitted afterward by a new `EmitWorkerThreadsReceiveMessageOnPortBody`, wired in right after `EmitMessageChannelTypes` and before `EmitRuntimeClassFinalize` (which closes `$Runtime`). The body synchronously dequeues one message from the port's own `_pending` queue and returns `{ message }` (a `Dictionary<string, object?>`), else `undefined` — matching `SharpTSMessagePort.ReceiveMessageSync`.

## 2. `messageerror` on `postMessage` of an uncloneable value (from #1001)

The emitted structured clone (`cloneCore`) returns uncloneable values by reference and never throws, so a compiled `$MessagePort` had no clone-failure point and always delivered `message`.

**Fix (confined to `$MessagePort`, per the issue's second option):** `PostMessage` detects the uncloneable categories up front (`typeof` `"function"` — covers functions/classes/bound-callable wrappers/delegates — and `"symbol"`) and enqueues a shared static `_cloneError` sentinel instead of a clone. `Drain` compares by reference and emits `'messageerror'` (no args) instead of `'message'`. This mirrors the interpreter's receiver-side model (`ClonedMessage(IsError: true)`) **without touching the shared `cloneCore`**, so compiled `structuredClone` / `$BroadcastChannel` are unaffected.

## Notes

- Both paths stay **fully standalone** — pure emitted IL, no `SharpTS.dll` dependency; verified with `--compile --verify` and `--compile --standalone --verify` (both clean, no SharpTS.dll co-located).
- Deeply nested uncloneables (e.g. `{ fn: () => {} }`) remain aliased, consistent with the emitted clone's existing shallow fallback — documented inline.
- Observed (pre-existing, out of scope): a compiled `$MessagePort` Refs the event loop on `Start()` unconditionally, unlike the interpreter which only Refs cross-thread ports. The existing compiled `MessageChannel` tests already work around this by closing ports; the new `messageerror` test follows that convention.

## Tests

- `ReceiveMessageOnPort_QueuedMessage_ReturnsMessageThenUndefined` — new, dual-mode (`{ message }` then `undefined`).
- `MessageChannelPort_PostUncloneable_FiresMessageError` — was an interpreter-only `[Fact]`, now a dual-mode `[Theory]`.

**Green:** xUnit 15428/0 · Test262 22/0/4skip (interp + compiled) · TSConf 31/0.

### DIFF
--- a/Compilation/RuntimeEmitter.MessageChannel.cs
+++ b/Compilation/RuntimeEmitter.MessageChannel.cs
@@ -32,6 +32,11 @@ public partial class RuntimeEmitter
     private FieldBuilder _messagePortClosedField = null!;
     private FieldBuilder _messagePortRefedField = null!;
     private FieldBuilder _messagePortOnEnqueueField = null!;
+    // Shared sentinel enqueued (in place of a clone) when postMessage receives an
+    // uncloneable value. Drain turns it into a 'messageerror' event, mirroring the
+    // interpreter's ClonedMessage(IsError: true) path (#1077). One instance per
+    // emitted assembly, created by $MessagePort's static ctor.
+    private FieldBuilder _messagePortCloneErrorField = null!;
     private ConstructorBuilder _messagePortCtor = null!;
     private MethodBuilder _messagePortDrain = null!;
     private MethodBuilder _messagePortStart = null!;
@@ -65,6 +70,18 @@ public partial class RuntimeEmitter
         // transferred to an interpreter worker, so a parent post wakes the worker loop
         // to drain _pending event-driven instead of the worker polling (#465).
         _messagePortOnEnqueueField = typeBuilder.DefineField("_onEnqueue", typeof(Action), FieldAttributes.Assembly);
+
+        // Static clone-failure sentinel (#1077). A single shared instance is enqueued in
+        // place of a message whose value cannot be structured-cloned; Drain compares by
+        // reference and emits 'messageerror' instead of 'message'.
+        _messagePortCloneErrorField = typeBuilder.DefineField(
+            "_cloneError",
+            _types.Object,
+            FieldAttributes.Assembly | FieldAttributes.Static | FieldAttributes.InitOnly);
+        var cctorIl = typeBuilder.DefineTypeInitializer().GetILGenerator();
+        cctorIl.Emit(OpCodes.Newobj, _types.GetDefaultConstructor(_types.Object));
+        cctorIl.Emit(OpCodes.Stsfld, _messagePortCloneErrorField);
+        cctorIl.Emit(OpCodes.Ret);
 
         EmitMessagePortConstructorIl(typeBuilder, runtime);
         EmitMessagePortDrain(typeBuilder, runtime);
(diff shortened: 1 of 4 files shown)


nickna:
Filed the two pre-existing divergences surfaced during this work as follow-ups:
- #1254 — compiled `$MessagePort.Start()` unconditionally Refs the event loop (started main-thread port hangs unless closed).
- #1255 — compiled structured clone (`cloneCore`) is shallow: aliases Date/RegExp/TypedArray/etc. and never throws `DataCloneError` on uncloneable values (this PR only works around the top-level `$MessagePort.postMessage` case).

Both are out of scope for #1077 and left as-is here.

PR #1458, opened 2026-08-24
Perf: add typed companions for stable primitive class methods

### LINKED ISSUE #1457: Perf: add typed companions for stable primitive class methods
Parent: #1278

## Problem

Merged PR #1456 (`ef4df4e8`) removed primitive field boxing and split the exact-class benchmark into direct field, method, and construction phases. The method phase is now the largest isolated residual:

| N=100,000 workload | SharpTS compiled | Node | Ratio |
|---|---:|---:|---:|
| reuse one exact `Counter` and call `step(): number` | 0.4904 ms | 0.0998 ms | **4.91x** |
| reuse one exact `Counter` through direct fields | 0.0545 ms | 0.0918 ms | 0.59x |
| construct exact instances | 0.7627 ms | 0.3866 ms | 1.97x |

BenchmarkDotNet at N=100,000 measured **1.248 ms / 2,400,121 B** for SharpTS method reuse, versus **176.7 us / 25 B** for idiomatic typed C# and **1.692 ms / 2,400,169 B** for the boxed-equivalent control. This is one boxed `double` per method result.

## Attribution

Synchronous public class methods are emitted as virtual CLR methods returning `object`, even when TypeScript proves a primitive return. Direct exact calls therefore invoke the virtual object ABI, box the method result, mark the IL stack unknown, then unbox/convert it for numeric consumers. Non-suspending direct calls also spill typed arguments through `object` locals before coercing them back to their declared CLR slots.

The object-returning virtual method remains necessary at observable boundaries: inheritance and overrides, extracted method values, reflection/runtime lookup, decorators, dynamic or computed dispatch, `super`, and uncertain receiver types.

## Proposed lowering

- Keep the existing public virtual object-returning method as the stable JavaScript ABI.
- For an eligible synchronous method with a proven total `number` or `boolean` return, emit one non-virtual typed companion containing the body.
- Make the public wrapper forward its arguments to the companion and box the primitive result once.
- At a proven exact and stable receiver call site, invoke the typed companion directly and retain its native stack type.
- Preserve typed arguments at eligible non-suspending calls instead of spilling them through `object`.
- Retain the ordinary virtual/dynamic path everywhere the proof is incomplete.

This avoids changing the public ABI or CLR override slots: a base-typed or dynamic call still reaches the most-derived virtual wrapper, whose own companion contains that override's body.

## Correctness boundary

Initial eligibility should be deliberately narrow:

- synchronous, non-generator, non-async instance methods;
- declared/inferred `number` or `boolean` result with no reachable bare return or fallthrough;
- no decorator, computed/symbol/private/accessor method, `arguments` capture, or generic/open dispatch;
- receiver proven to be the exact allocation from a stable local binding or an immediate `new C(...)` expression;
- method/prototype stability proof clear of reassignment, descriptor/prototype mutation, direct `eval`/`Function`, and opaque global access.

`this` on a potentially derived instance, `super`, base/interface-typed parameters, extracted methods, aliases without exact-allocation provenance, and every dynamic call retain the public wrapper path.

## Acceptance

- Permanent N=100,000 class-method cross-runtime median is no worse than **1.5x Node**.
- BenchmarkDotNet allocation is at most **1 KB/op** and time is at most **350 us/op** for the existing method-reuse workload.
- The selected hot loop contains a call to a primitive-returning companion and no result `box`, `unbox.any`, or generic numeric conversion.
- Public/extracted/dynamic calls still use the object-returning ABI.
- Base-typed calls dispatch to derived overrides; `super`, decorators, prototype mutation, and uncertain receivers preserve existing behavior.
- Standalone IL verification and the full compiler/conformance gates remain green.

## Follow-up

After this issue, exact construction remains the next measured target: **1.97x Node** and **10.68 MB/op** versus **2.29 MB/op** idiomatic C#, primarily from eager dynamic `_fields` allocation.

nickna:
Closes #1457.

## What changed

- emit an assembly-internal `double`/`bool` companion for eligible synchronous primitive-returning instance methods
- keep the public virtual object-returning method as the JavaScript/reflection/override ABI, forwarding to and boxing the companion result
- call the companion only for a unique exact `const x = new C(...)` receiver or immediate `new C().m()` shape
- reject aliases, escapes, captures, shadowing, mutable bindings, optional/default/rest parameters, `arguments`, decorators, generics, computed/private/async/generator methods, partial primitive returns, and uncertain/base-typed receivers
- retain the public wrapper whenever descriptors, `Object`, `eval`, `Function`, method reassignment, or class-prototype observation can make lookup observable

## Results

N=100,000, five independently launched paired runs:

| Runtime | Median |
|---|---:|
| SharpTS compiled | **0.0526 ms** |
| Node 24.19.0 | **0.0973 ms** |

The compiled result is 89.3% below the 0.4904 ms baseline and is now 0.54x Node.

BenchmarkDotNet ShortRun:

| Method | Mean | Allocated |
|---|---:|---:|
| Idiomatic C# | 168.6 us | 25 B |
| SharpTS | **307.8 us** | **114 B** |
| Boxed-equivalent C# | 1,381.5 us | 2,400,167 B |

The selected IL calls the primitive-returning companion directly and contains no result boxing; the public wrapper still boxes at observable boundaries.

## Validation

- Release solution build: green; standalone IL verification passed
- focused class/method/feature regression slice: 295/295 passed
- compiler test slice: 677 passed; three environment-only standalone failures (local TLS authentication and two child-process timeout classifications after the expected error was printed)
- focused BenchmarkDotNet and five paired cross-runtime launches completed successfully

### DIFF
--- a/src/SharpTS/Compilation/CompilationContext.Classes.cs
+++ b/src/SharpTS/Compilation/CompilationContext.Classes.cs
@@ -25,6 +25,14 @@ public partial class CompilationContext
     public bool IsInstanceMethod { get; set; }
 
     /// <summary>
+    /// Primitive-returning non-virtual method companions keyed by qualified class
+    /// and JavaScript method name. Only analyzer-marked exact receiver calls may
+    /// bypass the public virtual object-returning method.
+    /// </summary>
+    public Dictionary<string, Dictionary<string, MethodBuilder>>?
+        TypedPrimitiveInstanceMethodCores { get; set; }
+
+    /// <summary>
     /// True when the current method body is an inner <c>function</c> declaration
     /// emitted onto a display class. In that case <see cref="IsInstanceMethod"/> is
     /// also true (arg0 is the display-class self — needed for capture access), but
--- a/src/SharpTS/Compilation/ILCompiler.Classes.Methods.cs
+++ b/src/SharpTS/Compilation/ILCompiler.Classes.Methods.cs
@@ -412,6 +412,7 @@ public partial class ILCompiler
         returnType = _types.Object;
         if (classStatement.Decorators is { Count: > 0 }
             || classStatement.TypeParams is { Count: > 0 }
+            || classStatement.Fields.Any(field => field.IsDeclare)
             || method.Body is not { Count: > 0 } body
             || method.IsStatic
             || method.IsPrivate
--- a/src/SharpTS/Compilation/ILCompiler.Classes.Methods.cs
+++ b/src/SharpTS/Compilation/ILCompiler.Classes.Methods.cs
@@ -1,6 +1,8 @@
 using System.Reflection;
 using System.Reflection.Emit;
 using SharpTS.Parsing;
+using SharpTS.Parsing.Visitors;
+using SharpTS.TypeSystem;
 using TSTypeInfo = SharpTS.TypeSystem.TypeInfo;
 
 namespace SharpTS.Compilation;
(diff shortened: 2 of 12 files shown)


nickna:
Windows CI exposed one legitimate ambient-field edge case: a \\declare number\\ field is absent at runtime, so its first read must remain \\
ull\\ rather than coercing to \\

Issue #1599, opened 2026-09-07
Split EmittedRuntime metadata into cohesive feature components


nickna:
`EmittedRuntime` still holds substantial flat mutable metadata initialized with `null!`. Callers of that remaining surface must implicitly know which features and emission phases have initialized each handle. Optional features and forward declarations are legitimate requirements; this refactor makes their ownership and lifecycle explicit.

**Status: partially complete. Keep this issue open as the umbrella tracker for the remaining migrations.** The component pattern and five feature-family migrations have merged. This is continuation work using the established pattern.

Progress verified against `main` at [`b4c28152`](https://github.com/nickna/SharpTS/commit/b4c281525bdc2211c76b01fde7aeb9c94a2da630). See the [current holder](https://github.com/nickna/SharpTS/blob/b4c281525bdc2211c76b01fde7aeb9c94a2da630/src/SharpTS/Compilation/EmittedRuntime.cs#L18), [emission orchestration](https://github.com/nickna/SharpTS/blob/b4c281525bdc2211c76b01fde7aeb9c94a2da630/src/SharpTS/Compilation/RuntimeEmitter.cs#L37), and [documented migration pattern](https://github.com/nickna/SharpTS/blob/b4c281525bdc2211c76b01fde7aeb9c94a2da630/ARCHITECTURE.md#runtime-metadata-components). The original audit referenced [this earlier holder snapshot](https://github.com/nickna/SharpTS/blob/0b4f96592c1b9379de4ca5a52a2e8968af0fe365/src/SharpTS/Compilation/EmittedRuntime.cs#L18).

## Completed

- [x] Establish and document a consistent component lifecycle: explicit optional availability, checked declaration access, forward references before body emission, and a completion boundary that validates handles and rejects subsequent writes.
- [x] DNS: `EmittedDnsRuntime`, including the promise-wrapper registry. Merged in #1605.
- [x] Zlib: `EmittedZlibRuntime`, including compression and streaming handles. Merged in #1608.
- [x] TLS: `EmittedTlsRuntime`, including module/socket metadata and the deferred connect handle. Merged in #1610.
- [x] Datagram/UDP: `EmittedDgramRuntime`, including deferred receive and message-closure handles. Merged in #1620.
- [x] TCP/IPC networking: `EmittedNetRuntime`, including socket/server, factory, and BlockList metadata. Merged in #1624.

[...]

PR #1710, opened 2026-09-17
Own emitted function values and arguments metadata

nickna:
Function wrappers and argument objects currently expose twenty-four flat mutable handles. Give them required `FunctionValues` and `Arguments` owners, with completion checks at their original creation points. Eight emitters take scoped owners and exact peer declarations. Preserve the early argument-context field, later branded argument type, cache identities, invocation order and sixteen nullable-runtime paths.

Validation:
- Full-solution Release rebuild and source audit pass; all 102 existing compiler files preserve their original operations apart from metadata access. Thirteen test fixtures follow moved handles without changing assertions.
- Preview covers 9,061 distinct passing cases. Final integration passes 701 cases, including 35 new owner/invocation cases, nineteen isolated programs and the existing numeric allocation checks.
- Twenty-three saved assemblies preserve 25,860 IL-verified bodies: seventeen executable guests, two hosted outputs and four native configurations. Covers repeated emission, cache identity, receiver restoration, argument copies, strict/sloppy receivers and reference-assembly output.
- Quality checks retain the same 28 approved duplicate groups. One existing two-member async closure-emitter fingerprint follows the mechanical constructor-metadata move; its members and justification are unchanged.

The previously recorded console-restoration identity failure recurs in broad preview, while all twenty tests in its full class pass in preview and integration. It remains separately recorded; full CI must pass.

Part of #1599. Remaining binding/invocation/prototype metadata, iterators, dispatch/call expansion and the complete residual ownership/construction/registry/shared-infrastructure audit are still required.


nickna:
Superseded by smaller reviewable changes. CodeRabbit skipped this 121-file PR because its limit is 100 files; passing CI does not supply the missing review. #1711 contains the function-construction boundary in 67 files. Remaining function-value, argument and binding metadata is still tracked under #1599 and will follow after that chunk. This PR is closed unmerged; its source and verification evidence are retained.

PR #1812, opened 2026-09-19
refactor: localize Net construction metadata

nickna:
Net construction handles remain on the reusable emitter after their emission finishes. This change scopes 79 handles to one `EmitAll` call: socket/server declarations, six closure constructor/run pairs, BlockList helpers and local socket-error/drop-payload helpers.

Phase-one emitters return immutable construction values for early factories and later closure/body emission. TLS population receives the two Net fields it needs. The checked Net component remains the sole owner of public declarations. Optional availability, forward declarations, field visibility, method bodies and type-creation order stay unchanged.

Validation:
- Release build and unchanged code-quality/analyzer mutation gates passed.
- 5,947 regression cases passed, including all 116 focused cases and 1,386 standalone cases; the two existing Windows HTTP lifecycle skips remain.
- Two hosted/standalone reuse cases exercise earlier assemblies after Net, TLS, HTTP, full-runtime and minimal emissions, including factory options, independent BlockList state, IPC construction, half-close behavior and event-loop references.
- Nineteen IL-verified saved assemblies preserve normalized metadata, attributes, signatures and all 28,174 method bodies. Fourteen native profiles, three standalone executions and two hosted outputs pass against the baseline. BlockList, socket options and a loopback TCP exchange also match unchanged Node references.
- The source audit preserves 4,105 declaration/IL calls and emission order. Fresh committed-source binding confirms exactly 79 removed fields with no compilation, generator or binding errors.

Refs #1599. The candidate has zero flat declarations, 223 remaining construction fields and eight direct collections. The 34 TLS construction fields, required semantic corrections and complete final ownership audit remain in scope; this PR closes no final-audit rows.


nickna:
@coderabbitai review

PR #1813, opened 2026-09-19
refactor: localize TLS construction metadata

nickna:
TLS construction handles remain on the reusable emitter after emission finishes. This change scopes 37 handles to one `EmitAll` call: socket/server fields, handshake and accept closures, protocol/certificate helpers and certificate-identity helpers.

Phase-one emitters return immutable construction values for deferred bodies. Checked TLS declarations remain the source of socket/server types, and closure types come from their constructors. The TLS component now rejects duplicate declarations while allowing repair after failed completion. Feature selection, forward declarations, field visibility, Net field inputs and type-creation order stay unchanged.

Validation:

- Release build and unchanged code-quality/analyzer mutation gates passed.
- All 5,960 broad local regression cases passed on the preserved product/test binaries, including all 168 original focused cases and 1,386 standalone cases, with only two existing Windows HTTP lifecycle skips. After correcting six fixture hostname literals, the isolated Release test rebuild and all 170 focused/guard cases passed against the identical product compiler; original assertions and deadlines remain. No product source changed in that correction.
- Eleven lifecycle cases cover every public declaration, null/duplicate rejection, failed-completion repair and freezing. Two hosted/standalone reuse cases exercise earlier assemblies after TLS, minimal, Net, TLS+HTTP and full emissions, including certificate loading, SAN/identity checks, ALPN options and validation callbacks.
- Nineteen IL-verified saved assemblies preserve normalized metadata, attributes, signatures and all 27,711 method bodies. Fourteen native profiles pass on both compilers. Two standalone executions (identity checks and a real loopback TLS/ALPN exchange) match unchanged Node references; two hosted outputs compile and verify.
- The source audit preserves 1,555 declaration/IL calls and emission order. Fresh committed-source binding confirms exactly 37 removed fields, no added persistent metadata fields and no compilation, generator or binding errors.

The original named-import `new TLSSocket()` reference program compiles and verifies but fails before producing output on both baseline and candidate with `TypeError: not a constructor`. It is retained as required semantic work and excluded from passing execution counts. That guest never reaches `getProtocol()`; separately, saved native factories return null for an unconnected socket while the original Node null comparison returns false. Both observations remain required under #1599. The hosted constructor export was compiled and verified, not executed. Original sources, measured reference outputs, failures and deadlines are retained.

Refs #1599. The candidate has zero flat declarations, 186 remaining construction fields and eight direct collections. Required semantic corrections and the complete final ownership audit remain in scope; this PR closes no final-audit rows.

Initial CI failed only the source hermeticity guard on both platforms because the certificate fixture used `.example` labels. The correction uses `localhost`/`mismatch.localhost` and retains every assertion, test identity, trait and deadline. Corrected-head CI and incremental review remain required before merge.


nickna:
@coderabbitai review

nickna:
@coderabbitai review

PR #1841, opened 2026-09-20
Preserve stable immutable headers on fetched responses

nickna:
Fetched response headers were mutable in both runtimes, and the interpreter created a new Headers object on every property access. Fetched responses now expose a stable, immutable Headers object; clones retain a separate immutable object. Append, set, and delete throw a guest TypeError, including deletion of an absent header. Explicit Headers copies and constructed Request/Response headers remain mutable.

The emitted guard field belongs to the checked optional fetch implementation component and is required for completion. Mutation helpers receive the shared error component explicitly; no emitter-owned metadata cache was added. The interpreter caches its wrapper per fetched response.

Validation: 85 focused and lifecycle cases and 412 affected fetch, constructor, cookie, feature, and reflection cases passed. Release solution build, code-quality gates, and AOT analysis passed (zero analyzer warnings). The exact same saved source reproduces the baseline mismatch and matches Node after the fix on Windows and Linux, with standalone IL verification.

Refs #1599. This addresses fetched-header immutability and getter identity; the remaining semantic corrections and final residual-state audit remain open.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fetch response headers are now stable and immutable after a response is created.
  - Attempts to append, set, or delete headers on a fetched response now raise a `TypeError`.
  - Header copies made through `Headers`, `Request`, or `Response` remain mutable and do not alter the original response headers.

- **Tests**
  - Added coverage verifying header immutability, response isolation, and mutable copied headers across supported execution modes.


nickna:
@coderabbitai review

PR #1844, opened 2026-09-20
Refresh interface value bindings after declaration merging

nickna:
Values declared against an interface could retain its earlier shape after later declaration files augmented it. This made `Promise.withResolvers<number>()` fail even with ES2024 libraries selected, while a freshly declared `PromiseConstructor` value worked.

The checker now follows interface completion and merge replacements by object identity during value lookup. This replaces the Object-only name lookup workaround, preserves scope isolation when another interface shadows the same name, and resets the replacement tracking with declaration provenance at each checker entry point. Runtime Promise implementation and library selection remain unchanged: ES2023 still rejects this ES2024 API.

Tests cover ES2024/ES2025 Promise augmentation, ordinary cross-file ambient values, a shadowing namespace, invalid resolve arguments, and older-library rejection. All three initial regression cases failed before the fix; 140 focused cases and 4,470 broader type-checker/module/Promise feature and lifecycle cases pass. The unchanged saved source now matches Node (`true`, `7`) in the interpreter and isolated Windows/Linux standalone output with IL verification; the frozen baseline still rejects it. Code-quality gates and the zero-warning AOT baseline pass. The full Release solution build is verified before publication.

Refs #1599. This corrects the recorded rejection with explicitly selected ES2024 declarations. The compatibility default library remains ES5-rooted; selecting newer declarations is still necessary. Other semantic work and the complete residual-state audit remain open.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of interface augmentations from libraries and declaration files.
  - Ensured completed interface definitions remain correctly associated with their original declarations, even when similarly named types are in scope.
  - Improved support for APIs added through later library augmentations, including `Promise.withResolvers`.
  - Preserved accurate library selection, member typing, and diagnostics for unsupported APIs or invalid arguments.

- **Documentation**
  - Documented interface completion and declaration merging behavior during type checking.


nickna:
@coderabbitai review

nickna:
@coderabbitai review

PR #1845, opened 2026-09-20
Fix cooperative cancellation in emitted event loops

nickna:
## Problem and change

The early-emitted event loop had no cancellation helper, so flipping `$Runtime._cancelRequested` could not stop its callback drains. `WaitForTask` and `PumpOnce` also lacked checks within replenished queues, and a distant timer could postpone `Run`'s next poll.

Forward-declare `CheckCancellation` through the existing checked cancellation owner, then emit its body at the existing completion stage with a volatile flag read. All three entry points poll while draining callbacks, and `Run` caps timer waits at the existing 100 ms cancellation interval. The public flag/exception contract and standalone dependency rules remain unchanged.

Refs #1599; repairs the recorded event-loop omission against #74. Other semantic findings and the complete residual-state audit remain outstanding.

## Validation

- All four original hosted/plain Run/WaitForTask regression cases fail on the baseline and pass with the fix. Expanded coverage includes PumpOnce and a distant timer: 70 focused tests pass.
- 1,052 affected Promise, timer, event-loop, runtime-class and cancellation regressions pass with no skips.
- Saved baseline/candidate assemblies pass IL verification. A BCL-only runner verifies all three drains and the distant timer on Windows and Linux; baseline ignores cancellation, candidate stops after the first callback. Ordinary guest output is unchanged and neither saved assembly depends on SharpTS.dll.
- Full code-quality gates and analyzer mutation fixtures pass.
- Full Release solution build passes (34 existing NuGet warnings, zero errors); AOT/trim/single-file analyzer inventory remains at zero warnings.

CI exposed two existing declaration-order assertions. The cancellation signature now follows Stringify, FormatNumber and CreateException, preserving their original leading order. Both original assertions pass unchanged, with the repeated Windows/Linux comparisons and expanded regression suite passing.

## Summary by CodeRabbit

* **New Features**
  * Improved cooperative cancellation during event-loop execution.
  * Callback processing now checks for cancellation regularly, preventing continuously queued callbacks from delaying cancellation.
  * Long timer waits are limited to improve cancellation responsiveness.

* **Documentation**
  * Updated architecture documentation to describe cancellation handling across event-loop operations and runtime initialization.


nickna:
@coderabbitai review

